Issue #2, opened 2015-11-30
OpenShiftCreator assumes /oapi/v1 endpoints


warmi01:
OpenShiftCreator.java sets up the target URL endpoint using a /oapi/vi/ path string:

url = new URL(apiURL + "/oapi/v1/namespaces/" + namespace + "/" + path + "s");

However, when parsing a JSON List that contains a "Service" item, this undefined endpoint cannot be resolved (404) using /oapi/v1/namespaces/[namespace]/services (OpenShift).

The endpoint can be resolved using the Kubernetes (/api/v1) API endpoint: /api/v1/namespaces/[namespace]/services

Question..

Should there be a 'services' endpoint exposed in oapi/v1 or should this code use the api/v1 endpoint for "Service" requests?

gabemontero:
fixed via https://github.com/openshift/jenkins-plugin/commit/e720e3b6022bff59740dc51d7acdf91c53f8dd9d

PR #3, opened 2015-12-01
OpenShiftCreator assumes /oapi/v1 endpoints

warmi01:
Use Kubernetes API (/api/v1/) path for "services" requests (see open issue #2)

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftCreator.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftCreator.java
@@ -83,8 +83,8 @@ public class OpenShiftCreator extends Builder implements SimpleBuildStep, Serial
 		String response = null;
 		URL url = null;
     	try {
-    		if (chatty) listener.getLogger().println("\nOpenShiftCreator PUT URI " + "/oapi/v1/namespaces/"+namespace+"/"+path+"s");
-			url = new URL(apiURL + "/oapi/v1/namespaces/" + namespace + "/" + path + "s");
+    		if (chatty) listener.getLogger().println("\nOpenShiftCreator PUT URI " + (path.equals("service") ? "/api" : "/oapi") + "/v1/namespaces/"+namespace+"/"+path+"s");
+			url = new URL(apiURL + (path.equals("service") ? "/api" : "/oapi") + "/v1/namespaces/" + namespace + "/" + path + "s");
 		} catch (MalformedURLException e1) {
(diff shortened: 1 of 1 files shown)


gabemontero:
First - thanks for taking the initiative and looking to fix this.

Now, there is in fact a larger list than just services that falls under the "api" umbrella vs. "oapi" umbrella.  I'd rather address all the possibilities at once vs. just the one (i.e. services).

If you think you have the bandwidth to gather that list and update the PR, great.

If not, hopefully I'm going to dive into that research soon, but I've uncovered some other needed fixes based on recent master/slave testing that I need to sort out first.

If that keeps me occupied for too long, I'll reconsider merging in this subset fix in the interim.

thanks again

warmi01:
Re: "Now, there is in fact a larger list than just services that falls under the "api" umbrella vs. "oapi" umbrella. I'd rather address all the possibilities at once vs. just the one (i.e. services)."

Yup, that did occur to me (define mappings for all). I just figured for the sake of simplicity/clarity I would just target the offending case I ran into and report it to you, not knowing whether there were other considerations I wasn't aware of (e.g. new planned endpoints, REST API merge plans, etc).

Sure. I can take a crack at creating a more comprehensive fix.

Thanks for the quick turnaround.

gabemontero:
Cool deal - thx @warmi01 .  As a possible helpful pointer, I've heard (though have not confirmed with them directly) that the UI team may maintain a straight forward list of which API objects fall under api vs. oapi.  If so, perhaps @jwforres or @spadgett can confirm / elaborate.

spadgett:
@gabemontero @warmi01 You can see the list the web console uses here:

https://github.com/openshift/origin/blob/master/assets/app/config.js

gabemontero:
thanks @spadgett

warmi01:
thanks @spadgett

warmi01:
@spadgett -- so what we are actually looking for in this case are the subset of endpoints within the "namespaces" set of APIs, for example: /oapi/v1/namespaces/{namespace}/buildconfigs, /oapi/v1/namespaces/{namespace}/builds, ...), which is a somewhat lower level reduced set.

The lists you provided are 1st tier peer API endpoints with "/namespaces."

spadgett:
@warmi01 You might look at the swagger JSON to see which `oapi` resources fall under `/namespaces`, although unfortunately it's hard to read since it's raw JSON.

https://github.com/openshift/origin/blob/master/api/swagger-spec/oapi-v1.json

warmi01:
@spadgett Ah, yes. That works - helpful.

Thanks!

PR #4, opened 2015-12-03
Map resource "kinds" to API endpoints

warmi01:
Define mappings between resource "kinds" and associated API endpoints (i.e. OpenShift /oapi, Kubernetes /api).

Resource "kind" string is used as the key to apiMap for retrieving REST API type (oapi/api) and endpoint path.

Map uses a string array, which can provide for additional path fragments/parms, if needed later.

Note: "Template" maps to both "processedtemplates" (commented-out for now) and "templates" endpoints. So the desired one is not apparent based solely on the "Template" request kind.


warmi01:
Change applied via https://github.com/openshift/jenkins-plugin/commit/e720e3b6022bff59740dc51d7acdf91c53f8dd9d

PR #5, opened 2015-12-07
Fix coreLogic code path with null build parameter

warmi01:
When coreLogic is called with a null build parameter, inspectBuildEnvAndOverrideFields throws a NullPointer exception on build.getEnvironment(listener) call.


gabemontero:
Apologies ... didn't see this pull before commenting in the other pull.  Certainly feel free to run with this locally, but I think for an official fix, I want to leverage either the AbstractBuild or Run parameters based on the path into the build step in order to mine the env vars for overrides.

I'll most likely get to it on Tuesday, but if my description was clear enough, certainly feel free to move forward with it with this pull and beat me to the punch.

PR #12, opened 2016-01-15
Fix list formatting

rhuss:


gabemontero:
ah - looks better -  thx for that

PR #16, opened 2016-02-12
Add option to tag image using image reference

etiennecoutaud:
Hi, this PR is to add the possibility to tag an image with the OpenShiftImageTagger in ImageStreamImage mode using the reference of an image and not a tag.

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageTagger.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageTagger.java
@@ -21,21 +21,25 @@ import com.openshift.restclient.model.IImageStream;
 import javax.servlet.ServletException;
 
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.StringTokenizer;
 
 public class OpenShiftImageTagger extends OpenShiftBaseStep {
 
     protected String testTag = "origin-nodejs-sample:latest";
-    protected String prodTag = "origin-nodejs-sample:prod";    
+    protected String prodTag = "origin-nodejs-sample:prod";
+    private String tagImageReference = "false";
     
     // Fields in config.jelly must match the parameter names in the "DataBoundConstructor"
     @DataBoundConstructor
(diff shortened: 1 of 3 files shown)


gabemontero:
Hey @EtienneCoutaud 

So you'll need to rebase a bit.

First, on Thursday PM USA time, I checked in a change to split out the image stream name and tag name into separate fields / instance variables.  Your change looks like it is based on the old form.  So you'll need to resolve those conflicts.

Then after you opened this PR, based on input from the openshift-restclient-java folks the decision was made to go with [this PR](https://github.com/openshift/openshift-restclient-java/pull/113) instead of yours for the openshift-restclient-java change.  So you'll need to adjust this PR to consume the similar but not exactly the same API.

Lastly, I'm wondering if we can discern between ImageStreamImage and ImageStreamTag without having the user specify with a new config panel.  Here's what I'm thinking (though have not coded it up to confirm it is correct):
1) Call IImageStream.getImageId(String tagName) (on source image stream) with the parameter provided.
2) If it returns something, the tagName provided must be an ImageStreamTag
3) Otherwise, we can assume it is a ImageStreamImage
Can you try that out and see if it works?  If it does not work, we'll go with the new UI field you added.

thanks

etiennecoutaud:
Hi @gabemontero 

Okay I will try this using your change and the PR for the openshift-restclient-java to adjust with the new API.

Discerning between ImageStreamImage and ImageStreamTag without having the user specifiy is a good idea. I will implement that and see if it works

gabemontero:
Hey @EtienneCoutaud - if you don't think you'll get back to this, and are still interested, I can convert this to an issue for us to track and close out this PR.  We have reworked the tagging code a bit since you originally created this pull.   At the moment, you specify the tag name, and we translate it to a ImageStreamImage ref.  Longer term, I'm thinking we add a toggle to choose between ImageStreamImage and ImageStreamTag ref (i.e. the equivalent the `--alias` option).

gabemontero:
Hey @EtienneCoutaud - I've gone ahead and commited this enhancement via https://github.com/openshift/jenkins-plugin/commit/486f69bd22bfdab4ef72e6ba20ed0b682db61b2e, and as such, will close out this PR.  If you are at all still interested in this, please pull down the latest version and give it a try.

Thanks.

Issue #17, opened 2016-02-23
Better messages in Jenkins job logs


mfojtik:
>   BUILD STEP:  OpenShiftScaler in perform for frontend wanting to get to replica count 0 on namespace test

Should be something like "Scaling deployment 'frontend' in namespace 'test' to 0 replicas"

>   BUILD STEP EXIT:  OpenShiftScaler got the scale request through

Dunno what this means :-)

>   BUILD STEP:  OpenShiftBuilder in perform for ruby-sample-build on namespace test

Should be "Starting build #number for 'ruby-sample-build' in namespace 'test'". Alternatively, when the deployment check is in place we can add " and waiting for deployment to succeed".

>  BUILD STEP:  OpenShiftDeploymentVerifier in perform checking for frontend wanting to confirm we are at least at replica count 0 on namespace test

Should be "Waiting for the 'frontend' to reach number of replicas to be 0"

...
I think we need to cleanup this :-) I think the EXIT steps are not necessary. We should try to be less verbose and don't use something like "OpenShiftDeploymentVerifier" as users (and me) have no idea what that is :-) 

Also can we show links to OpenShift console so users can follow the build (to watch logs/cancel/etc)?

mfojtik:
@gabemontero @bparees what do you think?

bparees:
@mfojtik +1

@gabemontero and I already did some light reviewing of these messages, but there is definitely more aggressive editing that can be done.

gabemontero:
I'll add the specific message edits.  @bparees had another one he noted via a gist (though I'm not 100% sure what is wrong with what he posted, need to look again).  

And I think the idea of replacing the java class name with some sort of reference to what the build step is a good one.  I'm thinking the build step display name when they select it, unless you all have another suggestion.

Apologies, but I'm _not_ in agreement about losing the EXIT prints.  They have already served as useful debug, without the full verbose option, for customer questions, and have expedited either debug or correction to user error on several occasions already.

There is an option to dump the build output.  It currently is not real time because of some java level concerns.  The openshift-restclient-java package from the eclipse folks have real time version, but it employs the `oc` binary located in the jenkins image under the covers.  If you all recall, we made a conscious decision to not depend on the `oc`  binary (I'm assuming the reasons are self explanatory).

Links to the console ... is the login URL another env var that @mfojtik has recently added ? ... even if it is, I'm on the fence

bparees:
> Apologies, but I'm not in agreement about losing the EXIT prints

[...]

PR #23, opened 2016-03-02
new messages

gabemontero:
Fixes https://github.com/openshift/jenkins-plugin/issues/17

@bparees @mfojtik PTAL when you get the chance

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
@@ -1,0 +1,136 @@
+package com.openshift.jenkins.plugins.pipeline;
+
+public class MessageConstants {
+
+/*
+ * Generally speaking (exceptions will be specifically noted), we are going to use 
+ * %s String.format substitution 
+ * - the plugin name, when the message is shared across multiple plugins; otherwise, the name will baked into the message
+ * - state returned from the API server (i.e. a state of the deployment), unless we check for that explicit state, in which case it is baked into the message
+ * - the id for a build or deployment
+ * - when a specific input parameter is displayed	
+ */
+	
+// shared across all the plugins
+public static final String CANNOT_GET_CLIENT = "\n\nExiting \"%s\" unsuccessfully; a client connection to \"%s\" could not be obtained.";
+
+/*
+ * These messages are shared between "Trigger OpenShift Build" jenkins build step implemented by OpenShiftBuilder,
+ * "Verify OpenShift Builds" implemented by OpenShiftBuildVerifier, and "Cancel OpenShift Build(s)"
+ * 
+ */
+public static final String EXIT_BUILD_BAD = "\n\nExiting \"%s\" unsuccessfully; build \"%s\" has completed with status:  [%s].";
+public static final String EXIT_BUILD_GOOD_DEPLOY_BAD = "\n\nExiting \"%s\" unsuccessfully; build \"%s\" has completed with status:  [Complete]. However, not all deployments with ImageChange triggers based on this build's output triggered off of the new image.";
+public static final String EXIT_BUILD_GOOD_DEPLOY_IGNORED = "\n\nExiting \"%s\" successfully; build \"%s\" has completed with status:  [Complete].";
+public static final String EXIT_BUILD_GOOD_DEPLOY_GOOD = EXIT_BUILD_GOOD_DEPLOY_IGNORED + "  All deployments with ImageChange triggers based on this build's output triggered off of the new image.";
+public static final String START_BUILD_RELATED_PLUGINS = "\n\nStarting the \"%s\" step with build config \"%s\" from the project \"%s\".";
+
+/*
+ * These messages are for the "Trigger OpenShift Build" jenkins build step implemented by OpenShiftBuilder
+ */
+public static final String WAITING_ON_BUILD = "  Started build \"%s\" and waiting for build completion ...";
+public static final String WAITING_ON_BUILD_PLUS_DEPLOY = "  Started build \"%s\" and waiting for build completion followed by a new deployment ...";
+public static final String EXIT_BUILD_NO_BUILD_OBJ = "\n\nExiting \"" + OpenShiftBuilder.DISPLAY_NAME + "\" unsuccessfully; could not retrieve the associated Build object from the start build command.";
+public static final String EXIT_BUILD_NO_POD_OBJ = "\n\nExiting \"" + OpenShiftBuilder.DISPLAY_NAME + "\" unsuccessfully; the build pod for build \"%s\" was not found in time.";
+public static final String EXIT_BUILD_NO_BUILD_CONFIG_OBJ = "\n\nExiting \"" + OpenShiftBuilder.DISPLAY_NAME + "\" unsuccessfully; the build config \"%s\" could not be read.";
+
+/*
+ * These messages are for the "Verify OpenShift Builds" jenkins build step implemented by OpenShiftBuildVerifier
+ * Reminder - "Verify OpenShift Builds" looks at builds started externally from the Jenkins project this step is in 
+ */
+public static final String WAITING_ON_BUILD_STARTED_ELSEWHERE = "  Waiting on build \"%s\" to complete ...";
+public static final String WAITING_ON_BUILD_STARTED_ELSEWHERE_PLUS_DEPLOY = "  Waiting on build \"%s\" to complete followed by a new deployment ...";
+
+
+/*
+ * These messages are for the "Cancel OpenShift Builds" jenkins post-build action implemented by OpenShiftBuildCanceller
+ */
+public static final String CANCELLED_BUILD = "  Cancelled build \"%s\".";
+public static final String EXIT_BUILD_CANCEL = "\n\nExiting \"%s\" successfully with %d builds cancelled.";
+
+/*
+ * These messages are for the "Create OpenShift Resource(s)" jenkins build step implemented by OpenShiftCreator
+ */
+public static final String START_CREATE_OBJS = "\n\nStarting \"" + OpenShiftCreator.DISPLAY_NAME + "\" with the project \"%s\".";
+public static final String CREATED_OBJ = "  Created a \"%s\"";
+public static final String FAILED_OBJ = "  Failed to create a \"%s\"";
+public static final String EXIT_CREATE_BAD = "\n\nExiting \"" + OpenShiftCreator.DISPLAY_NAME + "\" unsuccessfully, with %d resource(s) created and %d failed attempt(s).";
+public static final String EXIT_CREATE_GOOD = "\n\nExiting \"" + OpenShiftCreator.DISPLAY_NAME + "\" successfully, with %d resource(s) created.";
+public static final String TYPE_NOT_SUPPORTED = "  The API resource \"%s\" is not currently supported by this step.";
+
+/*
+ * These messages are shared between "Cancel OpenShift Deployment", "Scale OpenShift Deployment", "Verify OpenShift Deployment", and 
+ * "Trigger OpenShift Deployment", implemented respectively by OpenShiftDeployCanceller, OpenShiftScaler, OpenShiftDeploymentVerifier, and OpenShiftDeployer
+ * 
+ */
+public static final String EXIT_DEPLOY_RELATED_PLUGINS_NO_CFG = "\n\nExiting \"%s\" unsuccessfully; the deployment config \"%s\" could not be read.";
+public static final String START_DEPLOY_RELATED_PLUGINS = "\n\nStarting \"%s\" with deployment config \"%s\" from the project \"%s\".";
+
+/*
+ * These messages are shared between "Verify OpenShift Deployment", and 
+ * "Trigger OpenShift Deployment", implemented respectively by OpenShiftDeploymentVerifier, and OpenShiftDeployer
+ * 
+ */
+public static final String EXIT_DEPLOY_RELATED_PLUGINS_BAD = "\n\nExiting \"%s\" unsuccessfully; deployment \"%s\" has completed with status:  [%s].";
+public static final String EXIT_DEPLOY_RELATED_PLUGINS_GOOD_REPLICAS_IGNORED = "\n\nExiting \"%s\" successfully; deployment \"%s\" has completed with status:  [Complete].";
+/*
+ * These messages are for the "Cancel OpenShift Deployment" jenkins post-build action implemented by OpenShiftDeployCanceller
+ * note we call this an "action" vs. "step" to help distinguish it as a post-build action vs. a build step
+ */
+public static final String EXIT_DEPLOY_CANCEL_GOOD_NOOP = "\n\nExiting \"" + OpenShiftDeployCanceller.DISPLAY_NAME + "\" successfully; the deployment \"%s\" is not in-progress; its status is:  [%s].";
+public static final String EXIT_DEPLOY_CANCEL_GOOD_DIDIT = "\n\nExiting \"" + OpenShiftDeployCanceller.DISPLAY_NAME + "\" successfully; the deployment \"%s\" has been cancelled.";
+public static final String EXIT_DEPLOY_CANCEL_BAD_NO_REPCTR = "\n\nExiting \"" + OpenShiftDeployCanceller.DISPLAY_NAME + "\" unsuccessfully; the latest deployment for \"%s\" could not be retrieved.";
+
+/*
+ * These messages are for the "Trigger OpenShift Deployment" jenkins build step implemented by OpenShiftDeployer
+ */
+public static final String EXIT_DEPLOY_TRIGGER_TIMED_OUT = "\n\nExiting \"" + OpenShiftDeployCanceller.DISPLAY_NAME + "\" unsuccessfully; gave up on deployment \"%s\" with status:  [%s].";
+
+/*
+ * These messages are for the "Verify OpenShift Deployment" jenkins build step implemented by OpenShiftDeploymenVerifier
+ */
+public static final String WAITING_ON_DEPLOY = "  Waiting on the latest deployment for \"%s\" to complete ...";
+public static final String WAITING_ON_DEPLOY_PLUS_REPLICAS = "  Waiting on the latest deployment for \"%s\" to complete and scale to \"%s\" replica(s) ...";
+public static final String EXIT_DEPLOY_VERIFY_GOOD_REPLICAS_GOOD = EXIT_DEPLOY_RELATED_PLUGINS_GOOD_REPLICAS_IGNORED + "  The deployment reached \"%s\" replica(s).";
+public static final String EXIT_DEPLOY_VERIFY_BAD_REPLICAS_BAD = EXIT_DEPLOY_RELATED_PLUGINS_BAD + "  And the replica count was \"%s\".";
+
+/*
+ * These messages are for the "Scale OpenShift Deployment" jenkins build step implemented by OpenShiftScaler
+ */
+public static final String SCALING = "  Scaling to \"%s\" replicas ...";
+public static final String SCALING_PLUS_REPLICA_CHECK = " Scaling to \"%s\" replicas and verifying the replica count is reached ...";
+public static final String EXIT_SCALING_NOOP = "\n\nExiting \"" + OpenShiftScaler.DISPLAY_NAME + "\" successfully; no deployments for \"%s\" were found, so a replica count of \"0\" already exists.";
+public static final String EXIT_SCALING_BAD = "\n\nExiting \"" + OpenShiftScaler.DISPLAY_NAME + "\" unsuccessfully; the call to \"%s\" failed.";
+public static final String EXIT_SCALING_TIMED_OUT = "\n\nExiting \"" + OpenShiftScaler.DISPLAY_NAME + "\" unsuccessfully; the deployment \"%s\" did not reach \"%s\" replica(s) in time.";
+public static final String EXIT_SCALING_GOOD = "\n\nExiting \"" + OpenShiftScaler.DISPLAY_NAME + "\" successfully for deployment \"%s\".";
+public static final String EXIT_SCALING_GOOD_REPLICAS_GOOD = "\n\nExiting \"" + OpenShiftScaler.DISPLAY_NAME + "\" successfully, where the deployment \"%s\" reached \"%s\" replica(s).";
+
+/*
+ * These messages are for the "OpenShift ImageStream" source code management plugin
+ * Note, there aren't clear cut start and exit points with this one; the SCM has multiple 
+ * entry points based on how Jenkins decides to call it
+ */
+public static final String SCM_CALC = "\n\nThe \"%s\" SCM will return the last revision state stored in Jenkins for the image stream \"%s\" and tag \"%s\" from the project \"%s\".";
+public static final String SCM_LAST_REV = "  Last revision:  [%s]";
+public static final String SCM_NO_REV = "  No revision state has been retrieved and stored yet.";
+public static final String SCM_COMP = "\n\nThe \"%s\" SCM is pulling the lastest revision state from OpenShift for the image stream \"%s\" and tag \"%s\" from the project \"%s\" and storing in Jenkins.";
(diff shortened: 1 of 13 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
+package com.openshift.jenkins.plugins.pipeline;
+
+public class MessageConstants {
+
+public static final String CANNOT_GET_CLIENT = "\n\nExiting \"%s\" unsuccessfully; a client connection to \"%s\" could not be obtained.";
+public static final String BAD_BUILD_STATE = "\n\nExiting \"%s\" unsuccessfully; build \"%s\" has completed with status:  [%s].";
+public static final String GOOD_BUILD_WITH_DEPLOY = "\n\nExiting \"%s\" successfully; build \"%s\" has completed with status:  [Complete].  All deployments with ImageChange triggers based on this build's output triggered off of the new image.";
+public static final String GOOD_BUILD_BAD_DEPLOY = "\n\nExiting \"%s\" unsuccessfully; build \"%s\" has completed with status:  [Complete]. However, not all deployments with ImageChange triggers based on this build's output triggered off of the new image.";
+public static final String GOOD_BUILD_IGNORE_DEPLOY = "\n\nExiting \"%s\" successfully; build \"%s\" has completed with status:  [Complete].";
+public static final String START_BUILD_CANCEL = "\n\nStarting the \"%s\" action for build config \"%s\" from the project \"%s\".";
```
bparees: why is this called start_build_cancel but the message is generic?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String CANCEL_SINGLE_BUILD = "  Cancelled build \"%s\".";
+public static final String EXIT_BUILD_CANCEL = "\n\nExiting \"%s\" successfully with %d builds cancelled.";
+public static final String START_BUILD = "\n\nStarting the \"%s\" step with build config \"%s\" from the project \"%s\".";
+public static final String BUILD_CANNOT_START = "\n\nExiting \"%s\" unsuccessfully; a build could not be started";
```
bparees: can we print the error here?  or will the error have been logged earlier?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String BUILD_STARTED = "  Started build \"%s\" and waiting for build completion %s...";
+public static final String BUILD_WAIT_DEPLOY = "followed by a new deployment";
+public static final String BUILD_NO_POD = "\n\nExiting \"%s\" unsuccessfully; the build pod for build \"%s\" was not found in time.";
+public static final String BUILD_NO_CFG = "\n\nExiting \"%s\" unsuccessfully; the build config \"%s\" could not be read.";
```
bparees: read? found?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String BUILD_VERIFYING = "  Verifying build \"%s\" and waiting for build completion %s...";
```
bparees: what are we verifying about the build?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String START_WITH_PROJECT = "\n\nStarting the \"%s\" step with the project \"%s\".";
+public static final String CREATING = "  Creating a \"%s\"...";
+public static final String TYPE_NOT_SUPPORTED = "  The API resource \"%s\" is not currently supported by this step.";
+public static final String START_DEPLOY_CANCEL = "\n\nStarting the \"%s\" action for deployment config \"%s\" from the project \"%s\".";
```
bparees: same question as build.  message name is cancel but the message is generic

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String DEPLOY_NO_CFG = "\n\nExiting \"%s\" unsuccessfully; the deployment config \"%s\" could not be retrieved.";
+public static final String DEPLOY_CANCEL_NOOP = "\n\nExiting \"%s\" successfully; the deployment \"%s\" is not in-progress; its status is:  [%s].";
+public static final String DEPLOY_CANCEL_SUCCESS = "\n\nExiting \"%s\" successfully; the deployment \"%s\" has been cancelled.";
+public static final String DEPLOY_NO_REP = "\n\nExiting \"%s\" unsuccessfully; the latest deployment \"%s\" could not be retrieved.";
+public static final String START_DEPLOY = "\n\nStarting the \"%s\" step with deployment config \"%s\" from the project \"%s\".";
```
bparees: same question for all the messages above this point i think.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String DEPLOY_FAILED = "\n\nExiting \"%s\" unsuccessfully; deployment \"%s\" has completed with status:  [Failed].";
+public static final String DEPLOY_TIMED_OUT = "\n\nExiting \"%s\" unsuccessfully; gave up on deployment \"%s\" with status:  [%s].";
+public static final String DEPLOY_SUCCESS = "\n\nExiting \"%s\" successfully; deployment \"%s\" has completed with status:  [Complete].";
+public static final String DEPLOY_VERIFY_STATE = "  Verifying deployment state%s...";
```
bparees: space after "state"?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String DEPLOY_VERIFY_REPLICAS = " and verifying the current deployment is at \"%s\" replica(s).";
+public static final String SCM_CALC = "\n\nThe \"%s\" SCM will return the last revision state stored in Jenkins for the image stream \"%s\" and tag \"%s\" from the project \"%s\".";
```
bparees, line 113: can we use another name other than SCM, or spell it out?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String SCM_LAST_REV = "  Last revision:  [%s]";
+public static final String SCM_NO_REV = "  No revision state has been retrieved and stored yet.";
+public static final String SCM_COMP = "\n\nThe \"%s\" SCM is pulling the lastest revision state from OpenShift for the image stream \"%s\" and tag \"%s\" from the project \"%s\" and storing in Jenkins.";
+public static final String SCM_CHANGE = "\n\n A revision change was found this polling cycle.";
+public static final String SCM_NO_CHANGE = "\n\n No revision change found this polling cycle.";
+public static final String START_TAG = "\n\nStarting the \"%s\" step with the source [image stream:tag] \"%s:%s\" and destination [image stream:tag] \"%s:%s\" from the project \"%s\".";
+public static final String EXIT_OK = "\n\nExiting \"%s\" successfully.";
+public static final String SCALING = "  Scaling to \"%s\" replicas%s...";
```
bparees: space after replicas

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String SCALING_VERIFYING = " and verifying the replica count is reached.";
+public static final String SCALING_NOOP = "\n\nExiting \"%s\" successfully; no deployments for \"%s\" were found, so a replica count of \"0\" already exists.";
+public static final String SCALING_FAILED = "\n\nExiting \"%s\" unsuccessfully; the call to \"%s\" failed.";
```
bparees: can we print the error or does it get logged earlier?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String SCALING_TIMED_OUT = "\n\nExiting \"%s\" unsuccessfully; the deployment \"%s\" did not reach \"%s\" replica(s) in time.";
+public static final String SCALING_SUCCESS = "\n\nExiting \"%s\" successfully%s.";
+public static final String SCALING_SUCCESS_REPLICAS = ", where the deployment \"%s\" reached \"%s\" replicas";
+public static final String SERVICE = "\n\nStarting the \"%s\" for the service \"%s\" from the project \"%s\".";
+public static final String SERVICE_CONNECTING = "  Attempting to connect to \"%s\"...";
+public static final String SERVICE_SUCCESS = "\n\nExiting \"%s\" successfully; a connection to \"%s\" was made.";
+public static final String SERVICE_FAILED = "\n\nExiting \"%s\" unsuccessfully; a connection to \"%s\" could not be made.";
+public static final String SOCKET_TO = " a socket level communication timeout to  \"%s\" occurred.";
```
bparees: there's an extra space after "to"

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,60 @@
...
+public static final String HTTP_ERR = " the HTTP level communication error \"%s\" to  \"%s\" occurred.";
```
bparees: extra space after "to"

bparees:
messages reviewed.  I didn't review the actual usage/invocation of the new constants so i'm going to trust/hope they are correct :)

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: ah ... the message is not as generic as it might imply (though the distinction may be too subtle);  the cancel build/deploy are "post-build actions" and the rest of them are "build steps" per jenkins UI terminology; so you see "action" in this message, and you see "step" in others.

I did consider using parameters for "action" vs. "step" and achieve some more message consolidation (there were also some opportunities to plugin in "deploy" vs. "build" for the config) ... gotta admin just got tired of string editing.

I can circle back and do that if you want.  Then the string constant name could be more generic as well.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: the openshift-restclient-java code does not currently expose the Build.Status.Reason field

if we really want to add that, I'd prefer doing it as part of a separate pull; if you want, I can open an issue to track that

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: the message corresponds to the openshift-restclient-java IClient.get returning null; that could mean the REST call got to the server and did not find it, or the call did not get through to the server ... hence I was hesitant to be precise and say "not found"

but I don't feel strongly ... shall I switch to "found" ?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: Verifying the build completes successfully, and that optionally that the appropriate deployments fired (the last %s will either be an empty string or a string re: the deployments).

How's this alternative for the message itself:  "  Waiting on build \"%s\" to complete %s..."

For the constant name  ... "WAIT_ON_BUILD" perhaps ?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: for the case where %s is the emtpy string, I would think we want "state..."

the non-empty string has a leading space

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero, line 113: yeah I struggled with this term ...I waffled on SCM (to the name of the section in the Jenkins UI panel) vs. something OpenShift specific like "ImageStream Poller"

Also, and this oddness has come up before, since we are broadening the interpretation of the term "source code" to now include images, do we want to say "Source Code Management"

"Version management" also came to mind, but then Jenkins doesn't really use that term.

I'm still somewhat undecided ... additional thoughts?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: ok ... so per the earlier comment ... you do want a space between the last word and the ... 

and yeah, after some searching, [the rule is spaces in between](http://www.thepunctuationguide.com/ellipses.html)

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: the only thing available is the message on the Exception that the openshift-restclient-java throws ... might be more harmful than good, so I omitted it

what do you think?

gabemontero:
wow ... you must really be busy down there :-)

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: I don't follow.  My question is why is this constant named "START_BUILD_CANCEL" and yet the action action for the message is a parameter, implying it is used for messages that may not relate to cancellatoin.

isn't this just a START_BUILD_ACTION message? (where the actual action is the first parameter argument)

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: I guess i'm not clear on under what conditions this message is thrown.  I was thinking you got an error back from the api call.  If this is a case where the build itself was successfully created and then entered some sort of error state (due to conditions in the master/cluster), then this is fine, the user can go inspect the build itself for more details.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: it's unfortunate we can't distinguish that, but given that limitation, read is probably the right call.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: yeah that sounds better.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: see my comments above, still unclear on this.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: all of these are still confusing as to why the first argument exists.  Can any of them be used with any other step than the one they are currently being used with?

eg would any parameter than "deployment cancel" make sense in this string: "Exiting [deployment cancel] successfully, the deployment [foo] has been cancelled" make sense?

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: got it, wasn't clear on how the %s was being used, thanks.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: so should the trailing space be removed from this string?  (the one before %s...)

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees, line 113: hm.  i guess let's leave it SCM for now.  My only other idea is to say "SCM imagestream poller"

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: well to be fair i just didn't understand how this was used w/ the other string, but based on your link and rethinking the aesthetics, yeah, might as well put the space there (so ignore my other comment about removing the space in one of the other strings)

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
bparees: i think we're going to want it when debugging.  is there any concept of verbose tracing in the jenkins config?

bparees:
@gabemontero did the review at the airport.  consumed my full 30mins of free wifi :(

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: in jenkins general, there is Logger tracing

for the plugin, yes, each step has a verbose switch that has trace points, etc.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: Been ruminating a bit on this since I saw your response last night.  Where I ended up was that my use of %s went off in the weeds a bit as I worried about consolidating strings and substituting in things like the step/action name (where that got tedious).

Where my mindset is now:
- have explicit strings for each step/action (but I'll still leverage constants for the step name)
- don't use %s for plugging in step/action name or for conditional checks (i.e. checking if deployments fired after the build completed)
- only use %s for dynamic information returned from the api server ... the state of something (failed or complete), or the name of a build or deployment

Couldn't tell for sure if case action vs. step was still a point of confusion.  Just in case:
- "step" is used for the selectable items from the "build step" drop down on the Jenkins UI
- "action" is used for the selectable items from the "post-build action" drop down

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero: a minor clarification came to mind:  replace "a build could not be started" with "could not obtain confirmation a build started" or "could not retrieve the associated Build object from the start build command " ... it more closely maps what is happening in the code, in that the API Build object was not returned.

bparees: +1

gabemontero:
OK ... reorganized the message file, with comments around the general methodology around use of %s, messages shared across the various plugin steps vs. those for a particular one, as well as addressed specific grammar/spacing/etc.

Hopefully better for the reviewer to consume.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,130 @@
...
+public static final String EXIT_OK = "\n\nExiting \"%s\" successfully.";
+
+/*
+ * These messages are for the "Verify OpenShift Service" jenkins build step implemented by OpenShiftServiceVerifier
+ */
+public static final String START_SERVICE_VERIFY = "\n\nStarting \"%s\" for the service \"%s\" from the project \"%s\".";
+public static final String SERVICE_CONNECTING = "  Attempting to connect to \"%s\" ...";
+public static final String EXIT_SERVICE_VERIFY_GOOD = "\n\nExiting \"%s\" successfully; a connection to \"%s\" was made.";
+public static final String EXIT_SERVICE_VERIFY_BAD = "\n\nExiting \"%s\" unsuccessfully; a connection to \"%s\" could not be made.";
+public static final String SOCKET_TO = " a socket level communication timeout to \"%s\" occurred.";
```
bparees: spell this out as TIMEOUT, took me a second to figure out "to what?" :)

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,130 @@
...
+
+/*
+ * These messages are shared between "Cancel OpenShift Deployment", "Scale OpenShift Deployment", "Verify OpenShift Deployment", and 
+ * "Trigger OpenShift Deployment", implemented respectively by OpenShiftDeployCanceller, OpenShiftScaler, OpenShiftDeploymentVerifier, and OpenShiftDeployer
+ * 
+ */
+public static final String EXIT_DEPLOY_RELATED_PLUGINS_NO_CFG = "\n\nExiting \"%s\" unsuccessfully; the deployment config \"%s\" could not be read.";
+public static final String START_DEPLOY_RELATED_PLUGINS = "\n\nStarting \"%s\" with deployment config \"%s\" from the project \"%s\".";
+public static final String EXIT_DEPLOY_RELATED_PLUGINS_BAD = "\n\nExiting \"%s\" unsuccessfully; deployment \"%s\" has completed with status:  [%s].";
+public static final String EXIT_DEPLOY_RELATED_PLUGINS_GOOD_REPLICAS_IGNORED = "\n\nExiting \"%s\" successfully; deployment \"%s\" has completed with status:  [Complete].";
```
bparees, line 75: These two exit messages that indicate the deployment status (and in one case indicates the deployment completed) don't seem like messages that would be used with a Cancel Deployment or Scale Deployment action, would they?

bparees:
i like the refactor.  1 nit and one question.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
gabemontero, line 75: Yep, the first start and exit are shared across those 4, but the two you noted are in fact only used in the "Trigger OpenShift Deployment" and "Verify OpenShift Deployment" steps.

How about I leave the constant names the same (since they apply to more than one step), but break them out into a separate section with a comment that just cites those two?

gabemontero:
Went ahead and pushed the change for the 1 nit plus my suggested response to the one question.

src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -0,0 +1,136 @@
...
+public static final String EXIT_OK = "\n\nExiting \"%s\" successfully.";
+
+/*
+ * These messages are for the "Verify OpenShift Service" jenkins build step implemented by OpenShiftServiceVerifier
+ */
+public static final String START_SERVICE_VERIFY = "\n\nStarting \"%s\" for the service \"%s\" from the project \"%s\".";
+public static final String SERVICE_CONNECTING = "  Attempting to connect to \"%s\" ...";
+public static final String EXIT_SERVICE_VERIFY_GOOD = "\n\nExiting \"%s\" successfully; a connection to \"%s\" was made.";
+public static final String EXIT_SERVICE_VERIFY_BAD = "\n\nExiting \"%s\" unsuccessfully; a connection to \"%s\" could not be made.";
+public static final String SOCKET_TO = " a socket level communication TIMEOUT to \"%s\" occurred.";
```
bparees: still says TO

gabemontero: Ah you meant the constant name - I thought you meant the message.  Will
switch.

bparees:
lgtm

gabemontero:
cool ... thx @bparees

PR #25, opened 2016-03-07
fix mvn compile issue upstream that does not show up locally

gabemontero:


gabemontero:
turns out this did not address the issue ... I had to modify the job at https://ci.openshift.redhat.com/jenkins/job/openshift-pipeline-plugin with these changes:
1) install java-1.8.0-openjdk-devel
2) set JAVA_HOME to /usr/lib/jvm/java-1.8.0-openjdk

PR #42, opened 2016-05-25
annotate builds and deployments with job info

bparees:


gabemontero:
igtm ... let's wait and see how the PR testing goes (though any breaks quite possibly will be on the testing end vs. these changes)

bparees:
@gabemontero still a WIP, i've added some more logic and some light refactoring.  but yes, let's see how this extended testing goes :)

bparees:
@spadgett fyi.  we should discuss what else you want/need annotated here.  (this is orthogonal to the service deps annotations which the jenkins plugin won't be responsible for)

gabemontero:
@bparees I see ... sounds good.  Btw, the extended test is waiting on the frontend and frontend-prod deployments as we type.

gabemontero:
@bparees at least with the first test, the frontend-prod deployment did not get kicked off (which is suppose to occur after the image tagging) ... the test job is essentially the old OpenShift Sample Job ... I'd be curious to see if you see the same problem if you created such a job in your sandbox.

gabemontero:
eesh ... the last test job failed on compilation with this:  `/usr/lib/golang/pkg/tool/linux_amd64/link: running gcc failed: fork/exec /bin/gcc: cannot allocate memory` while trying to build the OpenShift binaries for the test.

spadgett:
Thanks @bparees looks good.

Wonder if we should try to normalize on the URL value set between the sync and OpenShift plugins, particularly since we're using the same annotation key. The sync plugin will set a relative URI like `job/sample-pipeline/1` where this sets the full URL with scheme and host, correct? I think I prefer the full URL. Then we can make it a link in the UI without knowing the route.

spadgett:
@jwforres fyi

bparees:
>  The sync plugin will set a relative URI like job/sample-pipeline/1 where this sets the full URL with scheme and host, correct? 

yeah.  we can probably ask @jimmidyson to update the sync plugin to set the full url.  maybe open an issue against the sync plugin?

One problem w/ this implementation that hit me last night is the logic that will update a DC that was triggered by a BC, will only run if the "Verify deployment is triggered after build" option is selected when triggering a build.  If you don't select that option, we don't go down the path to find the RC that was created and update it w/ the annotation.  i'll see if i can fix that today.

bparees:
ok this is probably reviewable/mergeable so @spadgett has something to work with.

The major limitation right now is that when you use the plugin from a normal job, we can't get the full jenkins url, so the url will just be "job/job_name/job_number".  When it's used within a pipeline, we do get the full url, so in that case it'll be "http://jenkins.hostname.or.ip/job/job_name/job_number"

I can look into that as a follow up.

Things that are annotated with this change:
- any build triggered by the plugin
- any deployment(rc) triggered by the plugin
- any deployment(rc) triggered by the output of a build that was triggered by the plugin

bparees:
(merging it will also let @gabemontero get on w/ his refactoring)

src/main/java/com/openshift/jenkins/plugins/pipeline/Deployment.java
```diff
@@ -160,6 +168,8 @@ public static boolean didAllImagesChangeIfNeeded(String buildConfig, TaskListene
 		return true;
 	}
 
+	/* TODO: This appears to be dead code?
```
bparees: @gabemontero can you confirm this should be removed?

gabemontero:
Before looking at the changes, after initially at least running (though the frontend-prod deploy verify failed), the test job is currently getting an OOM compiling origin to run the extended tests.

@bparees - any ideas what is up?  I think the vagrant machine config parameters are the same as the origin/s2i analogous jobs.

And if it takes a while to sort it out, can you construct a test job in your local env that follows the classic jenkins job definition specified [here](https://github.com/openshift/origin/blob/master/test/extended/fixtures/testjob-plugin.xml) (it is pretty much the classic OpenShift Sample Job), and confirm it works for you.

thanks

src/main/java/com/openshift/jenkins/plugins/pipeline/Deployment.java
gabemontero: Yep, it is not called.  I must have replaced it with didAllImagesChangeIfNeeded and forgot to remove it.

bparees:
@gabemontero Not sure if this is what you're suggesting, but I did run a couple sample jenkins jobs using these plugin changes to confirm the annotations get added, during my dev process.

it would probably be good for me to extend your extended tests to have them actually check for the added annotations though.  Will consider that for a follow up.

src/main/java/com/openshift/jenkins/plugins/pipeline/Deployment.java
bparees: k. i've removed it.

src/main/java/com/openshift/jenkins/plugins/pipeline/AnnotationUtils.java
```diff
@@ -0,0 +1,59 @@
...
+			}
+			try { Thread.sleep(3000); } catch(InterruptedException e){}
+		}
+		if(!annotated) {
+			listener.getLogger().println(String.format(ANNOTATION_FAILURE, resource.getKind(), resource.getName()));
+		}
+		return annotated;
+	}
+
+}
```
gabemontero, line 59: Just an FYI ... I waffled back and forth between leaving this as a static/singleton and suggesting making a default method on IOpenShiftPlugin.

Let's leave it like this for now though.  I'll revisit when I merge this into my DSL induced refactor, and I can leverage eclipse as part of the analysis.

bparees, line 59: yeah, so i wasn't clear on why you'd made certain things defaults on IPlugin (you'll notice i also moved one of the other methods from IPlugin out to be a standalone static method).  Guessing i'm missing some context or intent.

gabemontero:
OK, sounds good.

Any ideas on the go build in the test job running out of memory?

src/main/java/com/openshift/jenkins/plugins/pipeline/AnnotationUtils.java
gabemontero, line 59: Yeah it was more of an organic evolution than coming up with explicit,
organizational thoughts.  The use of the default interface came after that
static Deployment class was created,
and I just stopped short of refactoring that class when I convered common
logic into IOpenShiftPlugin and OpenShiftBaseStep, etc.

As I said, I should revisit once I've merged your changes into my sandbox.
Thanks.

gabemontero:
Also, for future reference, you could follow the instructions [here](https://github.com/openshift/jenkins-plugin/tree/master/PR-Testing) and run the exact same extended tests locally.

bparees:
Nothing immediately comes to mind, but the AMIs recently updated to go1.6 which i believe is consuming more memory, so it's possible it's pushed things over some limit.

gabemontero:
Ok I'll recompare with the origin/s2i jobs see if memory was bumped up
somehow.  And I'll hit the merge button when I get in the office.

Thanks

gabemontero:
I've initiated a build at https://ci.openshift.redhat.com/jenkins/job/openshift-pipeline-plugin/ to build a version of the plugin with these changes.

I'll post the link to the plugin hpi file when it is complete so @spadgett can download it and give it a try.

gabemontero:
[Here it is](https://ci.openshift.redhat.com/jenkins/job/openshift-pipeline-plugin/lastSuccessfulBuild/artifact/plugin/openshift-pipeline.hpi)

src/main/java/com/openshift/jenkins/plugins/pipeline/Deployment.java
```diff
@@ -148,6 +149,11 @@ public static boolean didAllImagesChangeIfNeeded(String buildConfig, TaskListene
 			
 			if (didImageChangeFromPreviousVersion(client, dc.getLatestVersionNumber(), 
 					chatty, listener, dc.getName(), namespace, latestImageHexID, imageTag)) {
+
+				if(annotateRC) {
+					IReplicationController rc = getLatestReplicationController(dc, namespace, client, chatty ? listener : null);
+    				AnnotationUtils.AnnotateResource(client, listener, chatty, env, rc);
```
bparees, line 155: @gabemontero seems like this block might need to move as well to ensure we annotate the RC even if the image didn't change?

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftDeployer.java
```diff
@@ -98,10 +98,11 @@ public boolean coreLogic(Launcher launcher, TaskListener listener,
         			}
         			
     				try {
-    					rc = this.getLatestReplicationController(dc, client, overrides, chatty ? listener : null);
+    					rc = Deployment.getLatestReplicationController(dc, getNamespace(overrides), client, chatty ? listener : null);
     					if (chatty)
     						listener.getLogger().println("\nOpenShiftDeployer returned rep ctrl " + rc);
     					if (rc != null) {
+    	    				AnnotationUtils.AnnotateResource(client, listener, chatty, env, rc);
```
bparees, line 105: @gabemontero unless this annotation call will always ensure the RC is properly annotated upon being triggered by a deployment step?

src/main/java/com/openshift/jenkins/plugins/pipeline/Deployment.java
bparees: so we're only going to end up in this method now(after your change) if someone explicitly marked the build step to "verify deployment is triggered after build completes" which means they at least think they have an imagechangetrigger on the DC.

so i guess that means this is ok as is, we'll annotate deployments that were obviously triggered by the build (deployments for which the image changed), but only if they asked us to check for/verify the deployment explicitly.

and then the other block I referenced below will put the annotation on deployments that were directly triggered by a "deployment" job step.

so all in all I think we're ok once your change goes in.

PR #59, opened 2016-08-09
move to 5.0.0 snapshot, including new auth model with okhttp, resulti…

gabemontero:
…ng rework for create resources and dump logs; add cause to build request; switch to rest client pod logger for build logs (pod logger no longer dependent on oc binary)


gabemontero:
[test]

PR #65, opened 2016-09-26
WIP Added a menu for selecting wait-time units

oatmealraisin:
Added drop down menus to each step wait time.  These drop downs are for the units of the time.

If the user has previous values for the wait time, but no value for wait units, we retain the previous default and assume that the time is in milliseconds. Otherwise, the default unit is now seconds.

@bparees @gabemontero What do you both think?

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuildVerifier.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuildVerifier.java
@@ -1,4 +1,5 @@
 package com.openshift.jenkins.plugins.pipeline;
+import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftTimedStepDescriptor;
 import hudson.Extension;
 import hudson.util.FormValidation;
 import hudson.model.AbstractProject;
@@ -21,22 +22,23 @@ import java.util.Map;
 
 public class OpenShiftBuildVerifier extends OpenShiftBaseStep implements IOpenShiftBuildVerifier {
 	
-    protected final String bldCfg;
-    protected final String checkForTriggeredDeployments;
-    protected String waitTime; //TODO include in constructor / config.jelly
-    
-    // Fields in config.jelly must match the parameter names in the "DataBoundConstructor"
-    @DataBoundConstructor
-    public OpenShiftBuildVerifier(String apiURL, String bldCfg, String namespace, String authToken, String verbose, String checkForTriggeredDeployments) {
-    	super(apiURL, namespace, authToken, verbose);
-        this.bldCfg = bldCfg;
-        this.checkForTriggeredDeployments = checkForTriggeredDeployments;
-    }
+	protected final String bldCfg;
+	protected final String checkForTriggeredDeployments;
+	protected String waitTime; //TODO include in constructor / config.jelly
+	protected String waitUnit;
 
-    // generically speaking, Jenkins will always pass in non-null field values.  However, as we have periodically
-    // added new fields, jobs created with earlier versions of the plugin get null for the new fields.  Hence, 
-    // we have introduced the generic convention (even for fields that existed in the intial incarnations of the plugin)
-    // of insuring nulls are not returned for field getters
+	// Fields in config.jelly must match the parameter names in the "DataBoundConstructor"
+	@DataBoundConstructor
+	public OpenShiftBuildVerifier(String apiURL, String bldCfg, String namespace, String authToken, String verbose, String checkForTriggeredDeployments) {
+		super(apiURL, namespace, authToken, verbose);
+		this.bldCfg = bldCfg;
+		this.checkForTriggeredDeployments = checkForTriggeredDeployments;
+	}
+
+	// generically speaking, Jenkins will always pass in non-null field values.  However, as we have periodically
+	// added new fields, jobs created with earlier versions of the plugin get null for the new fields.  Hence,
+	// we have introduced the generic convention (even for fields that existed in the intial incarnations of the plugin)
+	// of insuring nulls are not returned for field getters
 
 	public String getBldCfg() {
 		return bldCfg;
@@ -49,112 +51,131 @@ public class OpenShiftBuildVerifier extends OpenShiftBaseStep implements IOpenSh
 	public String getWaitTime() {
 		return waitTime;
 	}
-	
+
+	public String getWaitUnit() {
+		return waitUnit;
+	}
+
 	public String getWaitTime(Map<String,String> overrides) {
 		String val = getOverride(getWaitTime(), overrides);
 		if (val.length() > 0)
-			return val;
-		return Long.toString(getDescriptor().getWait());
+			return val + checkWaitUnit(getWaitUnit());
+		return getDescriptor().getWait() + getDescriptor().getWaitUnit();
 	}
 	
 	
 	// Overridden for better type safety.
-    // If your plugin doesn't really define any property on Descriptor,
-    // you don't have to do this.
-    @Override
-    public DescriptorImpl getDescriptor() {
-        return (DescriptorImpl)super.getDescriptor();
-    }
+	// If your plugin doesn't really define any property on Descriptor,
+	// you don't have to do this.
+	@Override
+	public DescriptorImpl getDescriptor() {
+		return (DescriptorImpl)super.getDescriptor();
+	}
 
-    /**
-     * Descriptor for {@link OpenShiftBuildVerifier}. Used as a singleton.
-     * The class is marked as public so that it can be accessed from views.
-     *
-     */
-    @Extension // This indicates to Jenkins that this is an implementation of an extension point.
-    public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
-    	private long wait = GlobalConfig.BUILD_VERIFY_WAIT;
-        /**
-         * To persist global configuration information,
-         * simply store it in a field and call save().
-         *
-         * <p>
-         * If you don't want fields to be persisted, use <tt>transient</tt>.
-         */
+	/**
+	 * Descriptor for {@link OpenShiftBuildVerifier}. Used as a singleton.
+	 * The class is marked as public so that it can be accessed from views.
+	 *
+	 */
+	@Extension // This indicates to Jenkins that this is an implementation of an extension point.
+	public static final class DescriptorImpl extends BuildStepDescriptor<Builder>  implements IOpenShiftTimedStepDescriptor {
+		private String wait;
+		private String waitUnit;
+		/**
+		 * To persist global configuration information,
+		 * simply store it in a field and call save().
+		 *
+		 * <p>
+		 * If you don't want fields to be persisted, use <tt>transient</tt>.
+		 */
 
-        /**
-         * In order to load the persisted global configuration, you have to 
-         * call load() in the constructor.
-         */
-        public DescriptorImpl() {
-            load();
-        }
+		/**
+		 * In order to load the persisted global configuration, you have to
+		 * call load() in the constructor.
+		 */
+		public DescriptorImpl() {
+			load();
 
-        /**
-         * Performs on-the-fly validation of the various fields.
-         *
-         * @param value
-         *      This parameter receives the value that the user has typed.
-         * @return
-         *      Indicates the outcome of the validation. This is sent to the browser.
-         *      <p>
-         *      Note that returning {@link FormValidation#error(String)} does not
-         *      prevent the form from being saved. It just means that a message
-         *      will be displayed to the user. 
-         */
-        public FormValidation doCheckApiURL(@QueryParameter String value)
-                throws IOException, ServletException {
-        	return ParamVerify.doCheckApiURL(value);
-        }
+			doFillWaitUnit(GlobalConfig.getBuildVerifyWaitUnits());
+			doFillWait(GlobalConfig.getBuildVerifyWait());
+		}
 
-        public FormValidation doCheckBldCfg(@QueryParameter String value)
-                throws IOException, ServletException {
-        	return ParamVerify.doCheckBldCfg(value);
-        }
+		/**
+		 * Performs on-the-fly validation of the various fields.
+		 *
+		 * @param value
+		 *      This parameter receives the value that the user has typed.
+		 * @return
+		 *      Indicates the outcome of the validation. This is sent to the browser.
+		 *      <p>
+		 *      Note that returning {@link FormValidation#error(String)} does not
+		 *      prevent the form from being saved. It just means that a message
+		 *      will be displayed to the user.
+		 */
+		public FormValidation doCheckApiURL(@QueryParameter String value)
+				throws IOException, ServletException {
+			return ParamVerify.doCheckApiURL(value);
+		}
 
-        public FormValidation doCheckNamespace(@QueryParameter String value)
-                throws IOException, ServletException {
-        	return ParamVerify.doCheckNamespace(value);
-        }
-        
-        public FormValidation doCheckCheckForTriggeredDeployments(@QueryParameter String value)
-                throws IOException, ServletException {
-        	return ParamVerify.doCheckCheckForTriggeredDeployments(value);
-        }
-        
-        public FormValidation doCheckAuthToken(@QueryParameter String value)
-                throws IOException, ServletException {
-        	return ParamVerify.doCheckToken(value);
-        }
+		public FormValidation doCheckBldCfg(@QueryParameter String value)
+				throws IOException, ServletException {
+			return ParamVerify.doCheckBldCfg(value);
+		}
 
-        public boolean isApplicable(Class<? extends AbstractProject> aClass) {
-            // Indicates that this builder can be used with all kinds of project types 
-            return true;
-        }
+		public FormValidation doCheckNamespace(@QueryParameter String value)
+				throws IOException, ServletException {
+			return ParamVerify.doCheckNamespace(value);
+		}
 
-        /**
-         * This human readable name is used in the configuration screen.
-         */
-        public String getDisplayName() {
-            return DISPLAY_NAME;
-        }
-        
-        public long getWait() {
-        	return wait;
-        }
+		public FormValidation doCheckCheckForTriggeredDeployments(@QueryParameter String value)
+				throws IOException, ServletException {
+			return ParamVerify.doCheckCheckForTriggeredDeployments(value);
+		}
 
-        @Override
-        public boolean configure(StaplerRequest req, JSONObject formData) throws FormException {
-            // To persist global configuration information,
-            // pull info from formData, set appropriate instance field (which should have a getter), and call save().
-        	wait = formData.getLong("wait");
-        	GlobalConfig.setBuildVerifyWait(wait);
-            save();
-            return super.configure(req,formData);
-        }
+		public FormValidation doCheckAuthToken(@QueryParameter String value)
+				throws IOException, ServletException {
+			return ParamVerify.doCheckToken(value);
+		}
 
-    }
+		public boolean isApplicable(Class<? extends AbstractProject> aClass) {
+			// Indicates that this builder can be used with all kinds of project types
+			return true;
+		}
 
+		/**
+		 * This human readable name is used in the configuration screen.
+		 */
+		public String getDisplayName() {
+			return DISPLAY_NAME;
+		}
 
+		public String getWait() {
+			return wait;
+		}
+
+		public void setWait(String wait) {
+			this.wait = wait;
+		}
+
+		public String getWaitUnit() {
+			return waitUnit;
+		}
+
+		public void setWaitUnit(String waitUnit) {
+			this.waitUnit = waitUnit;
+		}
+
+		@Override
+		public boolean configure(StaplerRequest req, JSONObject formData) throws FormException {
+			// To persist global configuration information,
+			// pull info from formData, set appropriate instance field (which should have a getter), and call save().
+			wait = formData.getString("wait");
+			waitUnit = formData.getString("waitUnit");
+
+			GlobalConfig.setBuildVerifyWait(wait + waitUnit);
+			save();
+
+			return super.configure(req,formData);
+		}
+	}
 }
-
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
@@ -4,6 +4,7 @@ import java.util.Map;
 
 import javax.servlet.ServletException;
 
+import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftTimedStepDescriptor;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
@@ -20,36 +21,38 @@ import net.sf.json.JSONObject;
 
 
 public class OpenShiftBuilder extends OpenShiftBaseStep implements IOpenShiftBuilder {
-	
-    protected final String bldCfg;
-    protected final String commitID;
-    protected final String buildName;
-    protected final String showBuildLogs;
-    protected final String checkForTriggeredDeployments;
-    protected final String waitTime;
-    
-    
-    // Fields in config.jelly must match the parameter names in the "DataBoundConstructor"
-    @DataBoundConstructor
-    public OpenShiftBuilder(String apiURL, String bldCfg, String namespace, String authToken, String verbose, String commitID, String buildName, String showBuildLogs, String checkForTriggeredDeployments, String waitTime) {
-    	super(apiURL, namespace, authToken, verbose);
-        this.bldCfg = bldCfg;
-        this.commitID = commitID;
-        this.buildName = buildName;
-        this.showBuildLogs = showBuildLogs;
-        this.checkForTriggeredDeployments = checkForTriggeredDeployments;
-        this.waitTime = waitTime;
-    }
 
-    // generically speaking, Jenkins will always pass in non-null field values.  However, as we have periodically
-    // added new fields, jobs created with earlier versions of the plugin get null for the new fields.  Hence, 
-    // we have introduced the generic convention (even for fields that existed in the intial incarnations of the plugin)
-    // of insuring nulls are not returned for field getters
+	protected final String bldCfg;
+	protected final String commitID;
+	protected final String buildName;
+	protected final String showBuildLogs;
+	protected final String checkForTriggeredDeployments;
+	protected final String waitTime;
+	protected final String waitUnit;
+
+
+	// Fields in config.jelly must match the parameter names in the "DataBoundConstructor"
+	@DataBoundConstructor
+	public OpenShiftBuilder(String apiURL, String bldCfg, String namespace, String authToken, String verbose, String commitID, String buildName, String showBuildLogs, String checkForTriggeredDeployments, String waitTime, String waitUnit) {
+		super(apiURL, namespace, authToken, verbose);
+		this.bldCfg = bldCfg;
+		this.commitID = commitID;
+		this.buildName = buildName;
+		this.showBuildLogs = showBuildLogs;
+		this.checkForTriggeredDeployments = checkForTriggeredDeployments;
+		this.waitTime = waitTime;
+		this.waitUnit = waitUnit;
+	}
+
+	// generically speaking, Jenkins will always pass in non-null field values.  However, as we have periodically
+	// added new fields, jobs created with earlier versions of the plugin get null for the new fields.  Hence,
+	// we have introduced the generic convention (even for fields that existed in the initial incarnations of the plugin)
+	// of insuring nulls are not returned for field getters
 
 	public String getCommitID() {
 		return commitID;
 	}
-	
+
 	public String getBuildName() {
 		return buildName;
 	}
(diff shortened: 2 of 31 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
```diff
@@ -91,8 +93,27 @@ public String getWaitTime(Map<String,String> overrides) {
 			return val;
 		return Long.toString(GlobalConfig.getBuildWait());
 	}
-	
-	
+
+	public String getWaitUnits() { return this.waitUnit; }
+
+    @Override
+    public String getWaitTimeMilli(Map<String, String> overrides) {
```
gabemontero: You should be able to move this to IOpenShiftBuilder as a default method, and then share the implementation between the dsl and classic freestyle versions

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
```diff
@@ -69,15 +72,33 @@ public String getCheckForTriggeredDeployments() {
...
 		String val = getOverride(getWaitTime(), overrides);
 		if (val.length() > 0)
 			return val;
 		return Long.toString(getDescriptor().getWait());
 	}
-	
-	// Overridden for better type safety.
+
+    @Override
+    public String getWaitTimeMilli(Map<String, String> overrides) {
```
gabemontero: You should be able to move this to IOpenShiftBuilder as a default method, and then share the implementation between the dsl and classic freestyle versions

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
```diff
@@ -91,8 +93,27 @@ public String getWaitTime(Map<String,String> overrides) {
 			return val;
 		return Long.toString(GlobalConfig.getBuildWait());
 	}
-	
-	
+
+	public String getWaitUnits() { return this.waitUnit; }
+
```
gabemontero: Also, you'll need a setter for the wait units with an DataBoundSetter annotation for the wait units (those params get set via setters with the dsl, whereas they get set with the constructor for the classic freestyle jobs)

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
bparees: getWaitTimeMillis

src/main/java/com/openshift/jenkins/plugins/pipeline/model/GlobalConfig.java
```diff
@@ -27,6 +29,8 @@ public static long getBuildWait() {
 	public static void setBuildWait(long buildWait) {
 		GlobalConfig.buildWait = buildWait;
 	}
+	public static String getBuildWaitUnits() { return buildWaitUnits; }
+	public static void setBuildWaitUnits(String buildWaitUnits) { GlobalConfig.buildWaitUnits = buildWaitUnits; }
```
bparees: format these like the others, separate lines.

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
```diff
@@ -69,15 +72,33 @@ public String getCheckForTriggeredDeployments() {
...
+                break;
+        }
+
+        return Long.toString(duration);
+    }
+
+    // Overridden for better type safety.
     // If your plugin doesn't really define any property on Descriptor,
     // you don't have to do this.
     @Override
```
gabemontero: You'll also need to update the DescriptorImpl inner class to fully expose the default and capture whatever changes the user specifies for the global version of the setting.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/GlobalConfig.java
oatmealraisin: That's odd.. it shows up in intellij as one line.  Will change

bparees:
Can you not convert the stored value when displaying it to the user?  (and convert it back when saving it)?  

Having to continue to default the units to milliseconds (and force someone to select seconds from a drop down every time to change it) is painful.

gabemontero:
for the step level settings, setters should be added to the DSL impls; the constructor's serve as the setters for the classic job impls but if the user changes the units and saves, jenkins saves the updates on disk - the user does not have to change it everytime

For the global settings, the DescriptorImpl stuff I mentioned are the setter/ctors (where again Jenkins will persist config to disk); for the dsl stuff, the DescriptorImpl only handle step level changes (they don't really expose global level stuff for dsl steps) ... so the GlobalConfig class serves as a bridge for the DSL logic to access what was passed to the DescriptorImpl

bparees:
>  if the user changes the units and saves, jenkins saves the updates on disk - the user does not have to change it everytime

by change it every time i mean on every new field and new job.

gabemontero:
Ah i see - we could create a 4th value after milli, secs, and mins - a well
known named param which they can then override at the job or global level

bparees:
> Ah i see - we could create a 4th value after milli, secs, and mins - a well known named param which they can then override at the job or global level

having a global default for timeout units sounds pretty messy/complicated for users.

i'd prefer to see us just save the values in millis (because that's what we've already committed to) and display it in seconds and be done with it, assuming that is possible.

oatmealraisin:
@bparees @gabemontero Thanks for the clarification, I'll check if we can write back to the menu

gabemontero:
It is not possible from what I've seen but maybe there is some
ancient jelly taglib / JavaScript/ ajax gorp that can be put together.  Sam
might have some insight

oatmealraisin:
@gabemontero Sam?

gabemontero:
@spadgett was the Sam I was referring to @oatmealraisin

gabemontero:
btw, and with the acknowledgment that @oatmealraisin marked this WIP, there is another plugin PR, https://github.com/openshift/jenkins-plugin/pull/67, where the travis / test PR job ran successfully.

oatmealraisin:
@gabemontero @bparees I changed the default to 900 seconds by changing the order of the drop-down.   I don't think there is a way to write back to the rest of the text boxes, unfortunately.  I'm going to apply this to the rest of the boxes now.

bparees:
@oatmealraisin so if i'm a user who has a config where i set the value to 10,000milliseconds, when i update to this new plugin the value will be loaded and displayed as 10,000seconds?

oatmealraisin:
@bparees I see what you're asking for now, sorry.  Yes it would be displayed as 10000 seconds, but stored as 10000, with the units stored as milliseconds.  When getWaitTimeMillis is called, it calculates the millisecond value of the waitTime.  I didn't catch that earlier

bparees:
Sounds problematic.  Can we introduce units as part of the existing field value?

ie if the value is "1000" we assume it's milliseconds.  If it's "1000s" then we know it's seconds, "1000m" is minutes.

then you can allow users to select units (and display the correct unit value) while still supporting the existing configs?

oatmealraisin:
@bparees We could do that.  I think it may be hard to find out that we support it, though.  Currently I'm also looking at the last example of [this](https://wiki.jenkins-ci.org/display/JENKINS/Jelly+form+controls) to see if I can generate the dropdown with the current units at the top.

oatmealraisin:
@bparees @gabemontero I've implemented updating the drop down to reflect the user's previous selection.  I can still implement putting the units in with the amount if that would be better.

bparees:
> @bparees @gabemontero I've implemented updating the drop down to reflect the user's previous selection. I can still implement putting the units in with the amount if that would be better.

you still have to default the drop down to millis for this to work, right?  (since an existing config won't have a saved value for the drop-down, and the stored value will be in millis)

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
```diff
@@ -153,9 +166,38 @@ public Step newInstance(Map<String, Object> arguments) throws Exception {
...
+                    items.add("Minutes", "minutes");
+                    break;
+                case "milliseconds":
+                    items.add("Milliseconds", "milliseconds");
+                    items.add("Seconds", "seconds");
+                    items.add("Minutes", "minutes");
+                    break;
+            }
+            return items;
+        }
```
bparees: why is this duplicated?

oatmealraisin:
@bparees edit: You're right, I didn't think about previous configurations.  I'll change to using the inline unit notation, since that seems simpler at this point

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
oatmealraisin: Do you mean between the dsl/OpenShiftBuilder.java and pipeline/OpenShiftBuilder.java? I assumed since there were two Descriptors we needed the method in both.  I can remove it if I'm wrong

bparees:
@oatmealraisin well that's kinda the point...  you have to tolerate that the value you're loading might be an existing value a user created w/ the old version of the plugin, so you have to ensure it is respected for what it was intended to be.

So basically you need to treat all values as milliseconds in the absence of unit information.

You should be able to mutate the value when you're loading/saving it though.

So you load a config, the value is 1000, you show the user "1000" and "milliseconds"

the user changes the dropdown to "seconds" and saves it

you save "1000s"

next time the user loads the config, you show the user "1000" and "seconds"

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
bparees: there's got to be a way to share the code, no?

gabemontero: we can create an interface with a default method that this DescriptorImpl and the analogous ones for the other steps can implement (and then leverage

oatmealraisin:
@gabemontero @bparees I believe these changes should get the desired effect, ptal?  I've reverted most of the dsl class for now.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
```diff
@@ -41,7 +43,21 @@ default String getDisplayName() {
 	String getWaitTime();
 	
 	String getWaitTime(Map<String, String> overrides);
-	
+
+	// Used for if a user specifies a time unit with a wait time. Unitless numbers are considered milliseconds.
+	default long convertUnitNotation(String value) {
+		if(value.matches("\\d+milli")) {
+			return Long.valueOf(value.substring(0,value.length()-2));
```
bparees: -2? don't you need to strip the whole milli suffix?  similarly the other cases below.

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
```diff
@@ -92,7 +93,8 @@ public DescriptorImpl getDescriptor() {
      */
     @Extension // This indicates to Jenkins that this is an implementation of an extension point.
     public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
-    	private long wait = GlobalConfig.BUILD_WAIT;
+    	private String wait = GlobalConfig.getBuildWait();
+        private String waitUnit;
```
bparees: there's some random indenting throughout these changes

bparees (changes requested):

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
oatmealraisin: After discussion with Gabe, I've converted all the indentation in the files I worked in to tabs.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
oatmealraisin: Sorry, I missed that from when I changed from ms/s/m to milli/sec/min.  It's fixed now

oatmealraisin:
@gabemontero @bparees I've added all of the changes to the freestyle code.  ptal?  I still need to change the pipeline/dsl stuff

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuildVerifier.java
```diff
@@ -49,111 +51,175 @@ public String getCheckForTriggeredDeployments() {
 	public String getWaitTime() {
 		return waitTime;
 	}
+
+	public String getWaitUnit() {
+		if(waitUnit == null || waitUnit.length() == 0) {
+			if(waitTime.length() == 0) {
```
gabemontero: change to:

```
if (waitUnit  == null)
   return "milli";

if (waitUnit.length() == 0)
   return "sec";

return waitUnit;
```

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuildVerifier.java
```diff
@@ -49,111 +51,175 @@ public String getCheckForTriggeredDeployments() {
...
+		/**
+		 * In order to load the persisted global configuration, you have to
+		 * call load() in the constructor.
+		 */
+		public DescriptorImpl() {
+			load();
+
+			if(waitUnit == null || waitUnit.length() == 0) {
+				// If the user upgrades from an older version of the plugin, we need to set the units to milliseconds
+				if(wait != null && waitUnit.length() > 0) {
```
gabemontero: looks like you could hit a NPE here if you passed the outer loop cause waitUnit was null

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuildVerifier.java
```diff
@@ -49,111 +51,175 @@ public String getCheckForTriggeredDeployments() {
...
+			waitUnit = formData.getString("waitUnit");
+
+			GlobalConfig.setBuildVerifyWait(wait + waitUnit);
+			save();
+
+			return super.configure(req,formData);
+		}
+
+
+		// TODO: Find a place for this that multiple classes can use
```
gabemontero: I think you are clear on how to do this from prior comments / conversations, but since the comment was vague, just in case, it is "create an interface with a default method for doFillWaitUnitItems() that all the DescriptorImpl's processing wait units can implement"

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
```diff
@@ -20,165 +22,233 @@
...
 	}
-	
+
+	// NOTE: This is called by the the parameters in config.jelly
 	public String getWaitTime() {
 		return waitTime;
 	}
-	
+
+	public String getWaitUnit() {
```
gabemontero, line 76: same comment on changing the if's with the BuildVerifier module 

that said, this method should be moved to a default method within IOpenShiftPlugin, where you call the getters for waitUnit and wait, so that you only have to implement once

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftDeployer.java
```diff
@@ -46,105 +49,169 @@ public String getDepCfg() {
...
+		 */
+
+		/**
+		 * In order to load the persisted global configuration, you have to
+		 * call load() in the constructor.
+		 */
+		public DescriptorImpl() {
+			load();
+
+			if(waitUnit == null || waitUnit.length() == 0) {
```
gabemontero: the entire constructor after load could be moved into a default method in the interface all the DescriptorImpl's are going to implement that we've been talking about

gabemontero: With perhaps a generic method in GlobalConfig that takes a parameter like the outer class name so that it knows to call deploy wait units vs. build wait units

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuildVerifier.java
oatmealraisin: Yes, the todo was more to remind me than anything :)

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftDeployer.java
oatmealraisin: I created a default method w/ a string parameter, lmk if you'd like me to implement it more similarly to the way you suggested

oatmealraisin:
@gabemontero I added your suggestions, ptal?

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuildVerifier.java
```diff
@@ -49,111 +52,133 @@ public String getCheckForTriggeredDeployments() {
...
+		return (DescriptorImpl)super.getDescriptor();
+	}
+
+	/**
+	 * Descriptor for {@link OpenShiftBuildVerifier}. Used as a singleton.
+	 * The class is marked as public so that it can be accessed from views.
+	 *
+	 */
+	@Extension // This indicates to Jenkins that this is an implementation of an extension point.
+	public static final class DescriptorImpl extends BuildStepDescriptor<Builder>  implements TimedStepDescriptor {
```
gabemontero: we've bee using the naming convention of having all interfaces start with `IOpenShift`.  Convert `TimedStepDescriptor` to `IOpenShiftTimedStepDescriptor`

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftScalerPostAction.java
```diff
@@ -55,111 +57,173 @@ public String getVerifyReplicaCount() {
...
+		return (DescriptorImpl)super.getDescriptor();
+	}
+
+	/**
+	 * Descriptor for {@link OpenShiftScalerPostAction}. Used as a singleton.
+	 * The class is marked as public so that it can be accessed from views.
+	 *
+	 */
+	@Extension // This indicates to Jenkins that this is an implementation of an extension point.
+	public static final class DescriptorImpl extends BuildStepDescriptor<Publisher> {
```
gabemontero: can't this implement the new interface as well and remove some of the logic currently duplicated later in this class?

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
```diff
@@ -78,26 +80,34 @@ public String getCheckForTriggeredDeployments() {
...
+		return GlobalConfig.getBuildWait() + GlobalConfig.getBuildWaitUnits();
 	}
-	
-	
+
     private static final Logger LOGGER = Logger.getLogger(OpenShiftBuilder.class.getName());
 
 
 	@Extension
     public static class DescriptorImpl extends AbstractStepDescriptorImpl {
```
gabemontero: I believe you are reworking the DSL side of things now, so for now I'm assuming you will rework the DSL descriptor impl's similarly wrt interfaces with default methods, etc. ... I'll defer on further review of the DSL package until you tell me the DSL stuff should be looked at.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftTimedPlugin.java
```diff
@@ -0,0 +1,41 @@
...
+		}
+		if(value.matches("\\d+sec")) {
+			return Long.valueOf(value.substring(0,value.length()-3)) * 1000;
+		}
+		if(value.matches("\\d+min")) {
+			return Long.valueOf(value.substring(0,value.length()-3)) * 60 * 1000;
+		}
+		return Long.valueOf(value);
+	}
+}
```
gabemontero, line 38: yep / looks good

src/main/java/com/openshift/jenkins/plugins/pipeline/model/TimedStepDescriptor.java
```diff
@@ -0,0 +1,62 @@
+package com.openshift.jenkins.plugins.pipeline.model;
+
+import hudson.util.ListBoxModel;
+
+public interface TimedStepDescriptor {
```
gabemontero: add the `IOpenShift` as noted in the earlier comment

src/main/java/com/openshift/jenkins/plugins/pipeline/model/TimedStepDescriptor.java
```diff
@@ -0,0 +1,62 @@
...
+			default:
+				// We want the default units to be seconds, as opposed to milliseconds
+				items.add("Seconds", "sec");
+				items.add("Milliseconds", "milli");
+				items.add("Minutes", "min");
+		}
+
+		return items;
+	}
+}
```
gabemontero: after rename, yep / looks good

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftScalerPostAction.java
oatmealraisin: I had forgotten about editing this class.  Intellij is telling me that the class is never used, is that correct?

gabemontero:
Intellij is getting fooled by how Jenkins instropsects, etc.  It is getting
used by the config panel for the post build action.

oatmealraisin:
@gabemontero Should it have an entry in the GlobalConfig?

gabemontero:
It can reuse the scaler step's GlobalConfig

oatmealraisin:
@gabemontero I've added the dsl sections, ptal?

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
```diff
@@ -20,165 +21,192 @@
...
+
+	/**
+	 * Descriptor for {@link OpenShiftBuilder}. Used as a singleton.
+	 * The class is marked as public so that it can be accessed from views.
+	 *
+	 */
+	// NOTE: The getters in this class are used to populate the values of global.jelly, and the fields must match the
+	// method names
+	@Extension // This indicates to Jenkins that this is an implementation of an extension point.
+	public static final class DescriptorImplIOpenShift extends BuildStepDescriptor<Builder> implements IOpenShiftTimedStepDescriptor {
```
gabemontero: Interesting - I was always under the assumption that there was some naming restriction and that it had to be `DescriptorImpl`.  Good to know based on your testing that assumption is in fact incorrect.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
```diff
@@ -70,64 +72,107 @@ public String getWaitTime() {
...
+				Object waitTime = arguments.get("waitTime");
+				if (waitTime != null) {
+					wait = waitTime.toString();
+					step.setWaitTime(wait);
+				}
+			}
+
+			if (arguments.containsKey("waitUnit")) {
+				Object wUnit = arguments.get("waitUnit");
+				if (wUnit != null && wUnit.toString().length() != 0) {
```
gabemontero, line 145: shouldn't have to check if wUnit is null since you have the containsKey check above, right?

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
```diff
@@ -70,64 +72,107 @@ public String getWaitTime() {
...
+				throw new IllegalArgumentException("need to specify buildConfig");
+			OpenShiftBuildVerifier step = new OpenShiftBuildVerifier(bldCfg.toString());
+			if (arguments.containsKey("checkForTriggeredDeployments")) {
+				Object checkForTriggeredDeployments = arguments.get("checkForTriggeredDeployments");
+				if (checkForTriggeredDeployments != null) {
+					step.setCheckForTriggeredDeployments(checkForTriggeredDeployments.toString());
+				}
+			}
+
+			if (arguments.containsKey("waitTime")) {
```
gabemontero, line 135: move this if block to a default method on IOpenShiftTimedStepDescriptor

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
```diff
@@ -70,64 +72,107 @@ public String getWaitTime() {
...
+				Object waitTime = arguments.get("waitTime");
+				if (waitTime != null) {
+					wait = waitTime.toString();
+					step.setWaitTime(wait);
+				}
+			}
+
+			if (arguments.containsKey("waitUnit")) {
```
gabemontero, line 143: move this if block to a default method on IOpenShiftTimedStepDescritor

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
```diff
@@ -78,85 +81,127 @@ public String getCheckForTriggeredDeployments() {
...
+				}
+			}
+			if (arguments.containsKey("showBuildLogs")) {
+				Object showBuildLogs = arguments.get("showBuildLogs");
+				if (showBuildLogs != null) {
+					step.setShowBuildLogs(showBuildLogs.toString());
+				}
+			}
+
+			if (arguments.containsKey("waitTime")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftDeployer.java
```diff
@@ -69,47 +79,81 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+			if (!arguments.containsKey("deploymentConfig") && !arguments.containsKey("depCfg"))
+				throw new IllegalArgumentException("need to specify deploymentConfig");
+			Object depCfg = arguments.get("deploymentConfig");
+			if (depCfg == null || depCfg.toString().length() == 0)
+				depCfg = arguments.get("depCfg");
+			if (depCfg == null || depCfg.toString().length() == 0)
+				throw new IllegalArgumentException("need to specify deploymentConfig");
+			OpenShiftDeployer step = new OpenShiftDeployer(depCfg.toString());
+
+			if (arguments.containsKey("waitTime")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftDeployer.java
```diff
@@ -69,47 +79,81 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+				Object waitTime = arguments.get("waitTime");
+				if (waitTime != null) {
+					wait = waitTime.toString();
+					step.setWaitTime(wait);
+				}
+			}
+
+			if (arguments.containsKey("waitUnit")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
```diff
@@ -78,85 +81,127 @@ public String getCheckForTriggeredDeployments() {
...
+				Object waitTime = arguments.get("waitTime");
+				if (waitTime != null) {
+					wait = waitTime.toString();
+					step.setWaitTime(wait);
+				}
+			}
+
+			if (arguments.containsKey("waitUnit")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftDeploymentVerifier.java
```diff
@@ -87,57 +97,90 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+			if (!arguments.containsKey("deploymentConfig") && !arguments.containsKey("depCfg"))
+				throw new IllegalArgumentException("need to specify deploymentConfig");
+			Object depCfg = arguments.get("deploymentConfig");
+			if (depCfg == null || depCfg.toString().length() == 0)
+				depCfg = arguments.get("depCfg");
+			if (depCfg == null || depCfg.toString().length() == 0)
+				throw new IllegalArgumentException("need to specify deploymentConfig");
+			OpenShiftDeploymentVerifier step = new OpenShiftDeploymentVerifier(depCfg.toString());
+
+			if (arguments.containsKey("waitTime")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftDeploymentVerifier.java
```diff
@@ -87,57 +97,90 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+				Object waitTime = arguments.get("waitTime");
+				if (waitTime != null) {
+					wait = waitTime.toString();
+					step.setWaitTime(wait);
+				}
+			}
+
+			if (arguments.containsKey("waitUnit")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftScaler.java
```diff
@@ -84,55 +94,91 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+			if (depCfg == null || depCfg.toString().length() == 0)
+				depCfg = arguments.get("depCfg");
+			if (depCfg == null || depCfg.toString().length() == 0)
+				throw new IllegalArgumentException("need to specify deploymentConfig");
+			if (!arguments.containsKey("replicaCount"))
+					throw new IllegalArgumentException("need to specif replicaCount");
+			OpenShiftScaler step = new OpenShiftScaler(depCfg.toString(),
+					arguments.get("replicaCount").toString());
+
+			if (arguments.containsKey("waitTime")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftScaler.java
```diff
@@ -84,55 +94,91 @@ public BuildStepMonitor getRequiredMonitorService() {
...
+				Object waitTime = arguments.get("waitTime");
+				if (waitTime != null) {
+					wait = waitTime.toString();
+					step.setWaitTime(wait);
+				}
+			}
+
+			if (arguments.containsKey("waitUnit")) {
```
gabemontero: leverage the new default method on IOpenShiftTimedStepDescriptor called for earlier

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftTimedStepDescriptor.java
```diff
@@ -0,0 +1,62 @@
...
+
+	String getWait();
+
+	void setWait(String wait);
+
+	String getWaitUnit();
+
+	void setWaitUnit(String waitUnit);
+
+	default void doFillWaitUnit(String defaultWaitUnit) {
```
gabemontero, line 25: i think the dsl versions that i've called for being moved to this interface are different in signature than these existing methods, but that said, there could be the opportunity to share some logic - just keep that in mind

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftBuilder.java
oatmealraisin: This happened during a refactoring.  I thought I got all of the mistakes, I must have missed this one.  I've changed it back now, but it did work, which is interesting..

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
oatmealraisin: I'm not sure.. the other if blocks nearby lead me to believe so.

oatmealraisin:
@gabemontero I've implemented your changes, everything seems to be in order. ptal?

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
gabemontero: the `doFillWaitArguments` call I see in other step's `newInstance` is missing here ... was that intentional?

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
```diff
@@ -20,28 +19,24 @@
 import com.openshift.restclient.model.IBuildConfig;
 import com.openshift.restclient.model.IPod;
 
-public interface IOpenShiftBuilder extends IOpenShiftPlugin {
+public interface IOpenShiftBuilder extends IOpenShiftTimedPlugin {
```
gabemontero, line 22: ah ... yep, I like this new name better

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftTimedStepDescriptor.java
```diff
@@ -0,0 +1,82 @@
...
+			} else {
+				setWaitUnit(defaultWaitUnit);
+			}
+		}
+	}
+
+	default void doFillWaitArguments(Map<String, Object> arguments) {
+		if (arguments.containsKey("waitTime")) {
+			Object waitTime = arguments.get("waitTime");
+			if (waitTime != null) {
```
gabemontero, line 39: just call `doFillWait`

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftTimedStepDescriptor.java
```diff
@@ -0,0 +1,82 @@
...
+				setWait(waitTime.toString());
+			}
+		}
+
+		if (arguments.containsKey("waitUnit")) {
+			Object wUnit = arguments.get("waitUnit");
```
gabemontero, line 45: similarly seems like you could leverage `doFillWaitUnit` here

gabemontero (commented):
yep getting closer .... a few minor comments on the code; also, I realized that you should update the README.md with the new waitUnits parameter ... look for instances of waitTime

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
gabemontero: Good point :-) .... my comment stemmed from the fact that the containsKey was done just above, but I just double checked on stack overflow, and you can store nulls for both keys and values in maps.

Comment reverted, leave if check in :-)

bparees:
@oatmealraisin can you update the summary of this PR w/ an explanation of the new behavior and how existing configurations are handled?

oatmealraisin:
@bparees I've changed the top comment

@gabemontero @jupierce @bparees I'm closing this PR since Justin is merging it in #80

gabemontero:
@jupierce 's PR has merged.  Closing out per earlier comment from @oatmealraisin.

PR #68, opened 2016-09-27
Adding environment variable support to build request

jupierce:
In order to support environment variable feature request: https://trello.com/c/Yl4qomsw/878-5-additional-jenkins-plugin-enhancement-requests-from-github .

**Depends** on openshift/openshift-restclient-java#217

ptal @gabemontero @bparees 

Random changes:
- Global search and replace on "intial" misspell 
- Changed README.MD from using a) b) syntax to bullets to ease maintenance. 

DSL Syntax Example:
openshiftBuild( **env: [ test:'value', 'longenv1': 'env2']**, apiURL: 'https://10.13.137.85:8443' ....

Build Step Visualization:
buildstep

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/NameValuePair.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/NameValuePair.java
@@ -1,0 +1,40 @@
+package com.openshift.jenkins.plugins.pipeline;
+
+import hudson.Extension;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.Descriptor;
+import hudson.util.FormValidation;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
+
+public class NameValuePair extends AbstractDescribableImpl<NameValuePair> {
+
+    protected final String name;
+    protected final String value;
+
+    @DataBoundConstructor
+    public NameValuePair(String name, String value) {
+        this.name = name;
+        this.value = value;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends Descriptor<NameValuePair> {
+        public String getDisplayName() {
+            return "Name/Value Pair";
+        }
+
+        public FormValidation doCheckName(@QueryParameter String value) {
+            return FormValidation.validateRequired(value);
+        }
+    }
+
+}
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
@@ -1,4 +1,5 @@
 package com.openshift.jenkins.plugins.pipeline.dsl;
+import com.openshift.jenkins.plugins.pipeline.NameValuePair;
 import hudson.Extension;
 import hudson.model.Action;
 import hudson.model.BuildListener;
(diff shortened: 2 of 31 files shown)


jupierce:
Just a note that failure is expected until restclient is updated.

src/main/java/com/openshift/jenkins/plugins/pipeline/NameValuePair.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public String getValue() {
+        return value;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends Descriptor<NameValuePair> {
+        public String getDisplayName() {
+            return "Name/Value Pairs";
+        }
```
gabemontero, line 37: Can the user add a name value pair with empty string ... if so, for consistency sake, we could add checks either here or in the DescriptorImpl for OpenShiftBuilder that ensures something is at least set and we don't go down the env var path when we don't need to.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
```diff
@@ -153,6 +160,22 @@ public Step newInstance(Map<String, Object> arguments) throws Exception {
...
+            if ( envObject != null ) {
+                if ( envObject instanceof Map ) {
+                    Map map = (Map)envObject;
+                    Iterator<Map.Entry> it = map.entrySet().iterator();
+                    while ( it.hasNext() ) {
+                        Map.Entry entry = it.next();
+                        step.getEnvVars().add( new NameValuePair( entry.getKey().toString(), entry.getValue().toString() ) );
+                    }
+                } else {
+                    LOGGER.log(Level.WARNING, "Environment variables must be specified as a map. Received: " + envObject.getClass().getName() );
```
gabemontero: throwing an exception of some sort (i don't have strong preference between a runtime exception and a custom built exception) seems prudent here so that the error shows up in the job log; otherwise, the job may succeed without the env vars he was expecting and he may not think to double check the jenkins log

src/main/resources/com/openshift/jenkins/plugins/pipeline/NameValuePair/help-name.html
```diff
@@ -0,0 +1,3 @@
+<div>
```
gabemontero, line 1: can't tell from the git review tool ... did you link all the help files to the analogous locations for the dsl resources ... they are consumed by the pipeline grooy snippet generator thing

src/main/java/com/openshift/jenkins/plugins/pipeline/NameValuePair.java
jupierce: Will do. Thanks.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java
jupierce: Will change to UnsupportedOperationException.

src/main/resources/com/openshift/jenkins/plugins/pipeline/NameValuePair/help-name.html
jupierce: They are symlinked now.

jupierce:
@gabemontero Should be ready for another once over.

src/main/java/com/openshift/jenkins/plugins/pipeline/NameValuePair.java
gabemontero: Hey - I see where applyEnvVar trims and checks if the name is not empty before storing (don't remember if that was there before or not).  Was that your attempt to address this comment?

I was suggestion that you add validation methods to the DescriptorImpl for the name/value pair field (there are example in OpenShiftBuilder) and if you find issues, you can notify the user while he is constructing the build step.

src/main/resources/com/openshift/jenkins/plugins/pipeline/NameValuePair/help-name.html
gabemontero: Cool - thanks.

gabemontero:
Oh, and I see I forgot to say "thanks" for fixing misspellings and formatting in the readme :-)

PR #69, opened 2016-09-27
DSL step for ImageStreams SCM

csrwng:
Adds an openshiftImageStream DSL step to the OpenShift pipeline plugin.

Example:

```
openshiftImageStream name: 'myimage', namespace: 'myproject', tag: 'latest', verbose: 'true'
```

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
@@ -4,8 +4,12 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Map;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.servlet.ServletException;
 
+import hudson.EnvVars;
+import hudson.model.Job;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 
@@ -110,7 +114,7 @@ public class OpenShiftImageStreams extends SCM implements IOpenShiftPlugin {
     	IClient client = this.getClient(listener, DISPLAY_NAME, overrides);
     	
     	
-		IImageStream isImpl = client.get(ResourceKind.IMAGE_STREAM, getImageStreamName(overrides), namespace);
+		IImageStream isImpl = client.get(ResourceKind.IMAGE_STREAM, getImageStreamName(overrides), getNamespace(overrides));
 		// we will treat the OpenShiftImageStream "imageID" as the Jenkins "commitId"
 		String commitId = isImpl.getImageId(tag);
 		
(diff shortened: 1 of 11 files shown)


csrwng:
@gabemontero ptal

src/main/resources/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftImageStreams/help-apiURL.html
```diff
@@ -0,0 +1,6 @@
+<div>
```
gabemontero: can't tell from the review tool - do you create new files for all the different help files or link this file to the analogous file in the classic job package ...i've been able to link help files to avoid duplication

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
```diff
@@ -149,49 +153,66 @@ public SCMRevisionState calcRevisionsFromBuild(Run<?, ?> build,
...
 
-	@Override
-	protected PollingResult compareRemoteRevisionWith(
-			AbstractProject<?, ?> project, Launcher launcher,
-			FilePath workspace, TaskListener listener, SCMRevisionState baseline)
-			throws IOException, InterruptedException {
-    	listener.getLogger().println(String.format(MessageConstants.SCM_COMP, DISPLAY_NAME, getImageStreamName(project.getEnvironment(null, listener)), tag, namespace));
-    	String commitId = this.getCommitId(listener, project.getEnvironment(null,listener));
-		
+	protected PollingResult compareRemoteRevisionInternal(EnvVars env, Launcher launcher, TaskListener listener, SCMRevisionState baseline) {
```
gabemontero, line 172: Were you able to (re)confirm whether or not the k8s/openshift env vars like the project name, etc. were available in the env vars, in both the dsl/pipeline and classic/freestyle type jobs?

If available in either, being able to pull defaults would be cool.

src/main/resources/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftImageStreams/help-apiURL.html
csrwng: I did create new files. I will switch to links.

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
csrwng: So I tested different scenarios. When used within a pipeline, the PROJECT_NAME variable is not present. However, I created a regular build job with the ImageStreams SCM and the PROJECT_NAME was present in that case.

csrwng:
Changed copied files to links

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
gabemontero: OK cool/thanks.  Yeah, let's leave the UI pieces consistent and continue to share the common code where we don't attempt to leverage the PROJECT_NAME.

That said, the project name is also mounted in the file system at a well known location.  In hindsight I should have looked for it there (I do similar stuff with the token and cert in Auth.java).

We could add that as part of this pull, since we are mucking with this code path :-), and then not force the name space to be specified when setting the image stream pollers up.

Or if you'd rather just open an issue to handle this at a later time I'm good with that as well.

csrwng:
@gabemontero updated the ImageStreams SCM to get the namespace from wherever it's available.

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
```diff
@@ -226,12 +248,8 @@ public FormValidation doCheckApiURL(@QueryParameter String value)
...
         public FormValidation doCheckNamespace(@QueryParameter String value)
                 throws IOException, ServletException {
-        	// with some of the paths into the Image Stream SCM, the env vars typically available for the build steps are not available,
-        	// we we have to force the user to specify the project/namespace
-        	FormValidation fv = ParamVerify.doCheckNamespace(value);
-        	if (fv.kind == FormValidation.Kind.OK)
-        		return fv;
-        	return FormValidation.error("Please specify the name of the project");
+			// If a namespace is not specified, the default one is going to be used
+        	return ParamVerify.doCheckNamespace(value);
```
gabemontero, line 252: Ah - right, the IOpenShiftPlugin.getNamesapce(overrides) was already checking the file system.  So we simply had to relax the parameter verification - cool deal :-) !!

csrwng: Yeah, we were just not calling getNamespace(overrides) from the ImageStreams SCM. We are now.

gabemontero:
@csrwng I think the java side changes LGTM.

Now we should do some due diligence on programmatic testing for this.  In the past, I found automating testing for the IS polling tricky and have not tackled it to date.  That very well could still hold, but I'm curious on your take on it.

Are you familiar enough on where the jenkins extendedt test is / how it works, or would some pointers help?  After you've had a chance to digest them, we could huddle up and brainstorm a little.

csrwng:
@gabemontero should this be merged before adding the tests on the origin side?

gabemontero:
At first blush I could see it going either way (and was going to make the
call after our pow-wow),
at at second glance, sure, it is probably the path of least resistance to
merge first.

I will do that momentarily.

PR #70, opened 2016-10-04
Adding exec function to Jenkins OpenShift Plugin

jupierce:
Adding exec function for card:
https://trello.com/c/Yl4qomsw/878-5-additional-jenkins-plugin-enhancement-requests-from-github

Depends on openshift/openshift-restclient-java#220

ptal @gabemontero , @bparees 

Example verbose DSL:

```
       openshiftExec( apiURL: 'https://10.13.137.145:8443', 
        authToken: 'LsGHr_1XPhCkBVc7p6TivqVpvtDs0BDm9dIL8RdK0vU', 
        command: 'echo',
        arguments: [ [ value: 'hello'], [value: 'world'] ],
        namespace: 'myproject', 
        pod: 'ruby-hello-world-1-6939a', 
        verbose: 'true', 
        waitTime: '120000' 
       )
```

Example compact DSL (command and arguments in one groovy list):

```
       openshiftExec( apiURL: 'https://10.13.137.145:8443', 
         ...
        command: [ 'echo', 'hello', 'world' ],
       )
```

screenshot from 2016-10-04 11-43-07

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/Argument.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/Argument.java
@@ -1,0 +1,38 @@
+package com.openshift.jenkins.plugins.pipeline;
+
+import hudson.Extension;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.Descriptor;
+import hudson.util.FormValidation;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
+
+public class Argument extends AbstractDescribableImpl<Argument> {
+
+    protected final String value;
+
+    @DataBoundConstructor
+    public Argument(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+
+    @Extension
+    public static class DescriptorImpl extends Descriptor<Argument> {
+
+        @Override
+        public String getDisplayName() {
+            return "Argument";
+        }
+
+        public FormValidation doCheckValue(@QueryParameter String value) {
+            return FormValidation.validateRequired(value);
+        }
+
+    }
+
+}
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftExec.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftExec.java
@@ -1,0 +1,140 @@
+package com.openshift.jenkins.plugins.pipeline;
+
+import com.openshift.jenkins.plugins.pipeline.model.GlobalConfig;
+import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftApiObjHandler;
+import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftExec;
+import hudson.Extension;
+import hudson.model.AbstractProject;
+import hudson.tasks.BuildStepDescriptor;
+import hudson.tasks.Builder;
+import hudson.util.FormValidation;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
+import org.kohsuke.stapler.StaplerRequest;
+
+import javax.servlet.ServletException;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+public class OpenShiftExec extends OpenShiftBaseStep implements IOpenShiftExec, IOpenShiftApiObjHandler {
+
+    protected final String pod;
+    protected final String container;
+    protected final String command;
+    protected final List<Argument> arguments;
+    protected final String waitTime;
+
+    public String getPod() {
+        return pod;
+    }
+
+    public String getContainer() {
+        return container;
+    }
+
+    public String getCommand() {
+        return command;
+    }
+
+    public List<Argument> getArguments() {
+        return arguments;
+    }
+
+    public String getWaitTime() {
+        return waitTime;
+    }
+
+    @DataBoundConstructor
+    public OpenShiftExec(String apiURL, String namespace, String authToken, String verbose, String pod, String container, String command, List<Argument>arguments, String waitTime) {
+        super( apiURL, namespace, authToken, verbose );
+        this.pod = pod.trim();
+        this.container = container.trim();
+        this.command = command;
+        this.arguments = arguments;
+        this.waitTime = waitTime.trim();
+    }
+
+    // TODO: Copied from other class boilerplate, but I don't think this works. We are passing a user specified timeout string as a map key.
+    public String getWaitTime(Map<String,String> overrides) {
+        String val = getOverride(getWaitTime(), overrides);
+        if (val.length() > 0)
+            return val;
+        return Long.toString(GlobalConfig.getBuildWait());
+    }
+
+    @Extension // This indicates to Jenkins that this is an implementation of an extension point.
+    public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
+        private long wait = GlobalConfig.BUILD_WAIT;
+
+        /**
+         * In order to load the persisted global configuration, you have to
+         * call load() in the constructor.
+         */
+        public DescriptorImpl() {
+            load();
+        }
+
+        public FormValidation doCheckApiURL(@QueryParameter String value)
+                throws IOException, ServletException {
+            return ParamVerify.doCheckApiURL(value);
+        }
+
+        public FormValidation doCheckBldCfg(@QueryParameter String value)
+                throws IOException, ServletException {
+            return ParamVerify.doCheckBldCfg(value);
+        }
+
+        public FormValidation doCheckNamespace(@QueryParameter String value)
+                throws IOException, ServletException {
+            return ParamVerify.doCheckNamespace(value);
+        }
+
+        public FormValidation doCheckWaitTime(@QueryParameter String value)
+                throws IOException, ServletException {
+            return ParamVerify.doCheckCheckForWaitTime(value);
+        }
+
+        public FormValidation doCheckAuthToken(@QueryParameter String value)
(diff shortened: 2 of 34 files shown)


jupierce:
Note that test failure is expected until openshift/openshift-restclient-java#220 merges/installs.

jupierce:
Yep. src/main/resources/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec contains only symlinks.

src/main/java/com/openshift/jenkins/plugins/pipeline/Argument.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public String getValue() {
+        return value;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends Descriptor<Argument> {
+        public String getDisplayName() {
+            return "Command argument";
+        }
```
gabemontero, line 30: Can we do `FormValidation` style checking like we do in the build steps when extending Descriptor<Argument>?  If so, lets flag them if they add empty arguments.

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftExec.java
```diff
@@ -0,0 +1,130 @@
...
+        super( apiURL, namespace, authToken, verbose );
+        this.pod = pod.trim();
+        this.container = container.trim();
+        this.command = command;
+        this.arguments = arguments;
+        this.waitTime = waitTime.trim();
+    }
+
+    // TODO: Copied from other class boilerplate, but I don't think this works. We are passing a user specified timeout string as a map key.
+    public String getWaitTime(Map<String,String> overrides) {
```
gabemontero, line 60: @oatmealraisin 's PR is reworking this, creating a more common layer.  Depending on how merges first, some small refactor will be needed by the other PR to include this with the wait time stuff.

The map as returned by getOverride for wait time deals with the user using a Jenkins property in the config field, where at the top of the job, or at the global config panel, that property is assigned the number desired.

I'm pretty sure it worked originally when first introduced, but if you think you've found a bug, we can discuss.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec.java
```diff
@@ -0,0 +1,196 @@
...
+    protected String pod;
+    protected String container;
+    protected String command;
+    protected List<Argument> arguments = new ArrayList<>();
+    protected String waitTime;
+
+    @DataBoundConstructor
+    public OpenShiftExec( String pod ) {
+        this.pod = pod;
+    }
```
gabemontero, line 35: In the README you said the command was mandatory, but given the way the ctor is currently laid out, only the pod is mandatory.  At first blush, seems like the command should be mandatory and be part of the ctor.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec.java
```diff
@@ -0,0 +1,196 @@
...
+        private String argumentAsString(Map<String, Object> arguments, String arg ) throws IllegalArgumentException {
+            return requiredArgument(arguments, arg ).toString();
+        }
+
+        @Override
+        public Step newInstance(Map<String, Object> arguments) throws Exception {
+            OpenShiftExec step = new OpenShiftExec( argumentAsString( arguments, "pod" ) );
+            step.setContainer( argumentAsString( arguments, "container", "" ) );
+
+            Object commandObject = requiredArgument( arguments, "command" );
```
gabemontero, line 126: Yeah, you require it here (no default) ... command should be part of the ctor.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec.java
```diff
@@ -0,0 +1,196 @@
...
+        return container;
+    }
+
+    public String getCommand() {
+        return command;
+    }
+
+    public List<Argument> getArguments() {
+        return arguments;
+    }
```
gabemontero, line 71: Ahh ... separated the command from the arguments, since not strictly tied to java lang constructs here, unlike the restclient changes  ... +1

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		// get oc client
+		IClient client = this.getClient(listener, DISPLAY_NAME, overrides);
+		if ( client == null ) {
+			listener.getLogger().println(String.format(MessageConstants.EXIT_EXEC_BAD, DISPLAY_NAME, "Unable to obtain rest client"));
+			return false;
+		}
+
+		IPod p = client.get( ResourceKind.POD, getPod(), getNamespace());
```
gabemontero: I could see that if this step runs after a start deployment, and this step want to exec in the deployed pod, there could be delay on the openshift side where the deployed pod is ready, especially in environments like online or some of the test envs.  

I have encountered such behavior before.

Some form of loop and retry, where you punt if the wait interval is exceeded, like is done in some of the other steps, is warranted here.

src/main/resources/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec/config.jelly
```diff
@@ -0,0 +1 @@
+../../OpenShiftExec/config.jelly
```
gabemontero, line 1: ah - look at that - looks like the links were reflected here ... not sure why it didn't do that for the help files.  All good - just an observation.

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftExec.java
oatmealraisin: :+1:

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec.java
jupierce: @gabemontero I am allowing a couple compact forms of the DSL:
- `command : [ 'echo', 'hello', 'world'... ]`
- `command : 'echo', arguments : [ 'hello', 'world' ] ..`

In doing so, I can't rely on the default behavior to populate the constructor arguments correctly. I can remove these variants and rely on constructor, but the DSL syntax becomes extremely verbose.
`command : 'echo', arguments : [ [ value : 'hello' ], [ value : 'world' ] ], ..`

Let me know your preference.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec.java
jupierce: Please see note below on compact DSL forms for explanation.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec.java
gabemontero: Hmmmmmmmmmmmmmm :-) 

Just to make sure I understand, is it because only 1 constructor signature can have the @DataBoundConstructor annotation?

If so, then yep, I'm good with this exception.

src/main/java/com/openshift/jenkins/plugins/pipeline/Argument.java
jupierce: Adding in next push.

src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftExec.java
jupierce: My best assessment is that stapler does not support this scenario. You can have multiple DataBoundConstructors, but it is to add/deprecate parameters. It does not support the flexible unmarshalling this shorthand demands.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
jupierce: Added in next push.

gabemontero:
Seems plausible - let's go with what you've got.

jupierce:
@gabemontero PR test flake has been addressed.

jupierce:
Rebased.

jupierce:
Failure is a dependency on #75

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
```diff
@@ -0,0 +1,126 @@
...
+
+		// get oc client
+		IClient client = this.getClient(listener, DISPLAY_NAME, overrides);
+		if ( client == null ) {
+			listener.getLogger().println(String.format(MessageConstants.EXIT_EXEC_BAD, DISPLAY_NAME, "Unable to obtain rest client"));
+			return false;
+		}
+
+		IPod p = null;
+		for ( int retries = 10; p == null; retries-- ) {
```
gabemontero, line 64: I think we should factor in the wait time here vs. simply retries.  And then factor in whatever time we spend here if we have to retry into what is passed into `latch.await(..)`

gabemontero:
[test]

jupierce:
[test]

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
jupierce: Added,

gabemontero:
IGTM ... will hit the button after the test job completes.

PR #71, opened 2016-10-04
Adding support for multitag

jupierce:
Adding exec function for card:
https://trello.com/c/Yl4qomsw/878-5-additional-jenkins-plugin-enhancement-requests-from-github

ptal @gabemontero @bparees 

Enabling the following forms:
1. 1 destination stream and 1 destination tag   (i.e. single tag applies to single stream)
2. 1 destination stream and N destination tags   (i.e. all tags apply to the same stream)
3. 1 destination tag and N destination streams  (i.e. all streams will get the same tag)
4. N destination streams and N destination tags (i.e. each index will be combined to form a unique stream:tag combination)

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
@@ -1,25 +1,19 @@
 package com.openshift.jenkins.plugins.pipeline.model;
 
-import hudson.Launcher;
-import hudson.model.TaskListener;
-
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-
-import org.jboss.dmr.ModelNode;
-import org.jboss.dmr.ModelType;
-
 import com.openshift.internal.restclient.model.ImageStream;
 import com.openshift.jenkins.plugins.pipeline.Auth;
 import com.openshift.jenkins.plugins.pipeline.MessageConstants;
 import com.openshift.jenkins.plugins.pipeline.OpenShiftCreator;
 import com.openshift.restclient.IClient;
+import com.openshift.restclient.OpenShiftException;
 import com.openshift.restclient.ResourceKind;
-//import com.openshift.restclient.authorization.TokenAuthorizationStrategy;
 import com.openshift.restclient.model.IImageStream;
+import hudson.Launcher;
+import hudson.model.TaskListener;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+
+import java.util.*;
 
 public interface IOpenShiftImageTagger extends IOpenShiftPlugin {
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
```diff
@@ -132,113 +122,170 @@ default String deriveImageTag(String imageID, IImageStream srcIS) {
...
+				if ( iDestStreams.hasNext() ) {
+					destStream = iDestStreams.next().trim();
+				}
+
+				if ( iDestTags.hasNext() ) {
+					destTag = iDestTags.next().trim();
+				}
+
+				IImageStream destIS = null;
+				for ( int retries = 10; retries > 0; retries-- ) {
```
gabemontero, line 227: curious on the 10 retries here .... unlike some of the pod related operations discussed in the `oc exec` pull, I've not encountered similar timing issues with tagging images in the past.  Did it seem like it stemmed from the multi-tagging, where the code was creating multiple tags in rapid succession?

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
```diff
@@ -132,113 +122,170 @@ default String deriveImageTag(String imageID, IImageStream srcIS) {
...
+						isCreator.setAuth(Auth.createInstance(chatty ? listener : null, getApiURL(newOverrides), overrides));
+
+						boolean newISCreated = isCreator.coreLogic(launcher, listener, newOverrides);
+
+						if (!newISCreated) {
+							listener.getLogger().println(String.format(MessageConstants.EXIT_TAG_CANNOT_CREATE_DEST_IS, destStream, destNS));
+							return false;
+						}
+
+						for ( int retries2 = 10; retries2 > 0; retries2-- ) {
```
gabemontero: ah .. you just added this one while I was reviewing .... same curiosity on what exactly you are seeing in your testing and if the theory I posted in the comment above is on the right track or not

gabemontero (commented):
A couple of questions vs. direct comments, and yep, the deprecation of those methods was on my unofficial to do list

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
jupierce: Precisely that. Modifying the resource repeatedly caused a 403 unless I pulled the resource again.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
jupierce: That one was added more out of caution. I did not actually witness it fail. I'm fine to remove it if you think it unnecessary.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
gabemontero: Gotcha - thanks for the confirm.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
gabemontero: Yeah a get should be OK vs. the update ... let's remove / have a little less clutter.

gabemontero (commented):
Other than the one clean up item, I'm sure you did this, but just to be sure, did you test tagging where src and destination where in separate projects, where you added the necessary roles / tokens?

Otherwise, IGTM.

gabemontero:
Oh, just noticed the default test failure after my review comment.  We'll need to look into that as well prior to merging.

gabemontero:
Looks like the extended test set up failed @jupierce - see if this is a known flake or not in openshift/origin

jupierce:
@gabemontero I hand tested multi-project/multi-token configs. Looks good so far, but will need to implement extended tests to be sure those work on new 1->N and N->N variants.

gabemontero:
@jupierce roger that - sounds good.

gabemontero:
[test]

jupierce:
rebasing for conflict

jupierce:
@gabemontero PR test flake has been addressed.

gabemontero:
Cool - thanks for the confirm @jupierce

PR #77, opened 2016-10-06
Documentation suggestion 

jupierce:
- Moved common parameter help documentation to a single location. This eliminates dozens of help files that otherwise need to be independently kept in sync.
- Leveraged stapler's include function to render common parameters like cluster URL/time/verbose, etc. This should improve consistency and decrease boilerplace / maintenance throughout. 
- Found and fixed a few missing symlinks in the DSL config.jellys.

@gabemontero 

fwiw, I think this pattern could also help @oatmealraisin . The time control needs only to be implemented once in verbose-wait.jelly. The cluster test button would only need to be implemented once in cluster.jelly.


gabemontero:
Looks good!  Thanks!!!

I'll merge following the test runs.

oatmealraisin:
:tada:

jupierce:
@gabemontero Forgot to mention that I changed the authkey to a password field. Just wanted to make sure you were ok with that before  a merge.

gabemontero:
Nope, can't do that (migration issues).  Got to switch it back to the old
value.

gabemontero:
Sorry, misuderstood.  You didn't change the field name.  You changed the jelly field type.  My bad - that is fine.

PR #79, opened 2016-10-07
Refactoring exec doc to new pattern

jupierce:
The doc refactor merged before exec. Change to bring exec into alignment.


gabemontero:
IGTM - will do so when regression test finishes.

PR #78, opened 2016-10-07
Timeout fixes

jupierce:
Fixes:
- Several areas of the code were using the following to look up timeout values: `getOverrides(getWaitTime(), overrides ).` This ultimately looked up the user specified integer value in the overrides keymap. I don't think this was intentional.
- Several areas of the code were not using the GlobalConfig values. 
- All timeout value handling has been moved into a few classes for consistency and maintainability. 

ptal @gabemontero 

fwiw, this should also simplify @oatmealraisin 's pull for timeout units.

PR #80, opened 2016-10-07
Merge time units

jupierce:
Merging #65 on top of #78 

ptal @oatmealraisin @bparees @gabemontero

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBasePostAction.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBasePostAction.java
@@ -8,14 +8,20 @@ import java.io.Serializable;
 public abstract class TimedOpenShiftBasePostAction extends OpenShiftBasePostAction implements SimpleBuildStep, Serializable, ITimedOpenShiftPlugin {
 
     protected final String waitTime;
+    protected final String waitUnit;
 
-    public TimedOpenShiftBasePostAction(String apiURL, String namespace, String authToken, String verbose, String waitTime) {
-		super( apiURL, namespace, authToken, verbose );
-		this.waitTime = waitTime;
-	}
+    public TimedOpenShiftBasePostAction(String apiURL, String namespace, String authToken, String verbose, String waitTime, String waitUnit) {
+        super(apiURL, namespace, authToken, verbose);
+        this.waitTime = waitTime;
+        this.waitUnit = waitUnit;
+    }
 
-	final public String getWaitTime() {
-		return waitTime;
-	}
+    final public String getWaitTime() {
+        return waitTime;
+    }
+
+    final public String getWaitUnit() {
+        return TimeoutUnit.normalize(waitUnit);
+    }
 
 }
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBaseStep.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBaseStep.java
@@ -1,21 +1,34 @@
 package com.openshift.jenkins.plugins.pipeline;
 
+import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftPluginDescriptor;
 import com.openshift.jenkins.plugins.pipeline.model.ITimedOpenShiftPlugin;
+import hudson.model.AbstractProject;
+import hudson.model.Describable;
+import hudson.tasks.BuildStep;
+import hudson.tasks.BuildStepDescriptor;
 import jenkins.tasks.SimpleBuildStep;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.StaplerRequest;
 
 import java.io.Serializable;
 
-public abstract class TimedOpenShiftBaseStep extends OpenShiftBaseStep  implements SimpleBuildStep, Serializable, ITimedOpenShiftPlugin {
+public abstract class TimedOpenShiftBaseStep extends OpenShiftBaseStep implements SimpleBuildStep, Serializable, ITimedOpenShiftPlugin {
 
     protected final String waitTime;
+    protected final String waitUnit;
 
-    protected TimedOpenShiftBaseStep(String apiURL, String namespace, String authToken, String verbose, String waitTime) {
-		super( apiURL, namespace, authToken, verbose );
-		this.waitTime = waitTime;
+    protected TimedOpenShiftBaseStep(String apiURL, String namespace, String authToken, String verbose, String waitTime, String waitUnit) {
+        super(apiURL, namespace, authToken, verbose);
+        this.waitTime = waitTime;
+        this.waitUnit = waitUnit;
     }
 
     final public String getWaitTime() {
-		return waitTime;
-	}
+        return waitTime;
+    }
+
+    final public String getWaitUnit() {
+        return TimeoutUnit.normalize(waitUnit);
+    }
 
 }
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/GlobalConfig.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/GlobalConfig.java
@@ -1,71 +1,57 @@
 package com.openshift.jenkins.plugins.pipeline.model;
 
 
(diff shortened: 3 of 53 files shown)


jupierce:
Will finish dsl extension merge on Monday.

gabemontero:
looks great so far !

oatmealraisin:
@jupierce lgtm, I'm going to pull and test some stuff, thanks!

Edit: It looks like not everything is finished yet, I'll wait for the go-ahead on Monday! Have a nice weekend

src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBasePostAction.java
```diff
@@ -8,14 +8,20 @@
 public abstract class TimedOpenShiftBasePostAction extends OpenShiftBasePostAction implements SimpleBuildStep, Serializable, ITimedOpenShiftPlugin {
 
     protected final String waitTime;
+	protected final String waitUnit;
```
bparees: fix tabbing

src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBaseStep.java
```diff
@@ -8,14 +8,20 @@
 public abstract class TimedOpenShiftBaseStep extends OpenShiftBaseStep  implements SimpleBuildStep, Serializable, ITimedOpenShiftPlugin {
 
     protected final String waitTime;
+    protected final String waitUnit;
 
-    protected TimedOpenShiftBaseStep(String apiURL, String namespace, String authToken, String verbose, String waitTime) {
+    protected TimedOpenShiftBaseStep(String apiURL, String namespace, String authToken, String verbose, String waitTime, String waitUnit) {
 		super( apiURL, namespace, authToken, verbose );
 		this.waitTime = waitTime;
+        this.waitUnit = waitUnit;
```
bparees, line 23: tabbing

bparees: @gabemontero perhaps we should check in an eclipse format definition to the repo?  or at least describe one in the readme (tabs vs spaces, tab-stop settings, braces on same line or next line, etc)

src/main/java/com/openshift/jenkins/plugins/pipeline/model/ITimedOpenShiftPlugin.java
```diff
@@ -4,12 +4,47 @@
...
+
+	}
+
 	default long getDefaultWaitTime() {
 		return GlobalConfig.UBER_DEFAULT_WAIT;
 	}
 
 	String getWaitTime();
 
+    String getWaitUnit();
```
bparees, line 73: tabbing

bparees (changes requested):
couple minor formatting comments, otherwise lgtm so far.

gabemontero:
Yeah and we also have intellij vs eclipse going on here as well.  We can
confer at our next team meeting.  Maybe we need to go to make all tabs
spaces.

jupierce:
There is a mixture of 4space/tab throughout the files (even intra-file). Based on Gabe's comment, I will check in with 4space for now.

jupierce:
Sorry for the delay. I've found a bug in the original handling of global timeouts (i.e the currently public version of the jenkins plugin). The DSL implementations will only receive the configured global timeouts IFF the descriptor associated the pipeline build step is configured before the DSL step runs.
To recreate:
- Start a fresh Jenkins instance
- Configure the global wait time for builds to 1ms
- Create a pipeline job X with openshiftBuild DSL and do not specify timeout (set to empty string or do not specify at all).
- Run X
  - It should fail quickly because the 1ms timeout expires.
- Restart the Jenkins JVM  (http://...jenkinsurl.../restart)
- Directly after logging in, Run X
  - It will not fail and will use the static default timeout associated with builds.
  - This failure to use the configured timeout will persist until you save global configuration again.

I figured this was the right time to fix it since I'm in the code.

gabemontero:
On Mon, Oct 10, 2016 at 1:34 PM, Justin Pierce notifications@github.com
wrote:

> Sorry for the delay. I've found a bug in the original handling of global
> timeouts (i.e the currently public version). The DSL implementations will
> only receive the configured global timeouts IFF the descriptor associated
> the pipeline build step is configured before the DSL step runs.

Good catch!   Truth be told, this is not the first time the sequence of
something only getting set if you enter the configure screen has bitten me.

gabemontero:
Update: the port of @oatmealraisin 's time units change looks goo to me as well.

We've been discussing the format situation a bit on IRC.  I'm not going to halt this PR on that resolution.  Once we've decided (probably during a scrum) I'll go in and clean up all the java with the form we decide upon.

Otherwise, holding on merge to give @jupierce more time with the DSL difficulties around global config.  If the fix becomes a rabbit hole though, we'll punt on handling in this PR and merge this.

jupierce:
- Fixed pre-existing issues of global configuration failing to apply to DSL steps (this is actually the majority of the changes you will see).
- Fixed formatting issues called out in PR (but did not try to fix in all files).
- DSL waitUnits now incorporated from @oatmealraisin . 

All changes have been made to be backwards compatible.

jupierce:
To focus in on non-formatting changes, use the ?w=1 parameter to ignore whitespace: https://github.com/openshift/jenkins-plugin/pull/80/files?w=1

src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBasePostAction.java
jupierce: Fixed.

src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBaseStep.java
jupierce: Fixed.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/ITimedOpenShiftPlugin.java
jupierce: Fixed.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/GlobalConfig.java
```diff
@@ -1,71 +1,57 @@
...
-	private static long buildVerifyWait = DEFAULT_BUILD_VERIFY_WAIT;
-	private static long deployWait = DEFAULT_DEPLOY_WAIT;
-	private static long deployVerifyWait = DEFAULT_DEPLOY_VERIFY_WAIT;
-	private static long scalerWait = DEFAULT_SCALER_WAIT;
-	private static long execWait = DEFAULT_EXEC_WAIT;
+    public static final long DEFAULT_EXEC_WAIT = 3 * MINUTE;
 
-	private static int serviceVerifyRetry = DEFAULT_SERVICE_VERIFY_RETRY;
+    public static long getBuildWait() {
+        return new OpenShiftBuilder.DescriptorImpl().getConfiguredDefaultWaitTime();
```
gabemontero, line 29: A comment somewhere ... maybe the abstract descriptor class you created, would be good here, as I think we've fallen into the realm of Jenkins black magic again.  

If I'm reading the tea leaves correctly, but instantiating the DescriptorImpl, you are forcing a load of the associated global config that was saved previously to occur.  Is that it?

(Is so, very clever diagnosis / solution hacking :-) ... was this solution even googable ?)

gabemontero (commented):
Just one question / request for comment on the global config for dsl fix.

Pending that update and completion of the regression job IGTM.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/GlobalConfig.java
jupierce: @gabemontero Added a few comments in TimedBuildStepDescriptor.java . Your interpretation is spot on. I could eliminate the black magic if we were willing to forego backwards compatibility for the global wait times, but I will leave that decision to the team.

Google had no roads to this particular destination :-)

PR #83, opened 2016-10-11
Added a "Test Connection" button to each build step

oatmealraisin:
These check for a basic "Server Ready" status on the given URL,
skipping TLS authentification

I've abandoned the idea of having global "server profiles" for now, due to time. 

@gabemontero Unfortunately, extending IOpenShiftPlugin requires our descriptors to implement coreLogic(), which I don't think is desirable.  I'll reach out to you to discuss possibly splitting IOpenShiftPlugin into a few more classes, so that we can implement only what we need.

@gabemontero @bparees ptal?

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/HttpUtils.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/HttpUtils.java
@@ -1,0 +1,111 @@
+package com.openshift.jenkins.plugins.pipeline.model;
+
+import com.openshift.internal.restclient.DefaultClient;
+import com.openshift.internal.restclient.authorization.AuthorizationContext;
+import com.openshift.internal.restclient.okhttp.OpenShiftAuthenticator;
+import com.openshift.internal.restclient.okhttp.ResponseCodeInterceptor;
+import com.openshift.jenkins.plugins.pipeline.Auth;
+import com.openshift.restclient.http.IHttpConstants;
+import com.openshift.restclient.utils.SSLUtils;
+import hudson.model.TaskListener;
+import okhttp3.Dispatcher;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.X509TrustManager;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+public class HttpUtils {
+    static String httpGet(boolean chatty, TaskListener listener, Map<String, String> overrides, String urlString, String authToken, String displayName, DefaultClient client, Auth auth, String apiURL) {
+        URL url = null;
+        try {
+            url = new URL(urlString);
+        } catch (MalformedURLException e) {
+            if(chatty)
+                e.printStackTrace(listener.getLogger());
+            return null;
+        }
+        // our lower level openshift-restclient-java usage here is not agreeable with the TrustManager maintained there,
+        // so we set up our own trust manager like we used to do in order to verify the server cert
+        try {
+            AuthorizationContext authContext = new AuthorizationContext(authToken, null, null);
+            ResponseCodeInterceptor responseCodeInterceptor = new ResponseCodeInterceptor();
+            OpenShiftAuthenticator authenticator = new OpenShiftAuthenticator();
+            Dispatcher dispatcher = new Dispatcher();
+            OkHttpClient.Builder builder = new OkHttpClient.Builder()
+                    .addInterceptor(responseCodeInterceptor)
+                    .authenticator(authenticator)
+                    .dispatcher(dispatcher)
+                    .readTimeout(IHttpConstants.DEFAULT_READ_TIMEOUT, TimeUnit.MILLISECONDS)
+                    .writeTimeout(IHttpConstants.DEFAULT_READ_TIMEOUT, TimeUnit.MILLISECONDS)
+                    .connectTimeout(IHttpConstants.DEFAULT_READ_TIMEOUT, TimeUnit.MILLISECONDS)
+                    .hostnameVerifier(auth);
+            X509TrustManager trustManager = null;
+            if (auth.useCert()) {
+                trustManager = Auth.createLocalTrustStore(auth, apiURL);
+            } else {
+                // so okhttp is a bit of a PITA when it comes to enforcing skip tls behavior
+                // (it should just allow you to set a null ssl socket factory given how
+                // RealConnection/Address/Route work), but stack overflow came to the rescue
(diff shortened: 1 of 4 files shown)


gabemontero:
@oatmealraisin actually, I think it is better to extend IOpenShiftPlugin and provide a no-op implementation for coreLogic().  That way, you leverage all the existing code for connecting via the httpGet method and we can support TLS vs. no TLS in the same way like we discussed yesterday.

An agreeable splitting up / refactoring IOpenShiftPlugin to avoid the coreLogic no-op would be the best of both worlds, but I'm fine with deferring on that to get this in this week.

So apologies, but please switch back to the other approach we discussed.

gabemontero:
I think I'm OK with the abandoing server profile piece.  Still thinking about whether the new cluster.jelly @jupierce recently added is the right spot for this.  But we can circle back to this after you refactor around IOpenShiftPlugin.  I would think it is reasonably straight forward to move the UI widget between jelly files.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -37,5 +43,30 @@ default ListBoxModel doFillWaitUnitItems() {
         return items;
     }
 
+    default FormValidation doTestConnection(@QueryParameter String apiURL, @QueryParameter String authToken) {
+        Map<String, String> fakeOverride = new TreeMap<String, String>();
+        fakeOverride.put("SKIP_TLS", "true");
```
bparees: as discussed offline, we definitely can't deliver this w/ skip_tls=true

gabemontero: ok cool we are in agreement

oatmealraisin:
@gabemontero I still had trouble implementing IOpenShiftPlugin in this interface, so instead I just implemented httpGet. ptal?

gabemontero:
@oatmealraisin can we either:
1) create an interface that both IOpenShiftPlugin and IOpenShiftPluginDescriptor extend that just has httpGet as a default method that shares the actual code / avoids code duplication where ever possible
2) create a static class that the httpGet in both interfaces leverages (again share code, etc. as much as possible)

At first blush 1) is my first choice.

oatmealraisin:
@gabemontero I've changed httpGet to a static method, and also changed its signature to take vars it would normally get through getters.  ptal?

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPlugin.java
```diff
@@ -624,8 +624,7 @@ default boolean annotateJobInfoToResource(IClient client, TaskListener listener,
         return annotated;
     }
 
-    default String httpGet(boolean chatty, TaskListener listener,
-                           Map<String, String> overrides, String urlString) {
+    static String httpGet(boolean chatty, TaskListener listener, Map<String, String> overrides, String urlString, String authToken, String displayName, DefaultClient client, Auth auth, String apiURL) {
```
gabemontero: Let's create a separate class called `HttpUtils` in the model package and put this static method there.

Then leave `httpGet` as a default method in `IOpenShiftPlugin`, where in that method you call `HttpUtils.httpGet`

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftApiObjHandler.java
```diff
@@ -20,7 +21,8 @@
 	
 	default String fetchApiJsonFromApiServer(boolean chatty, TaskListener listener, 
     		Map<String,String> overrides, String apiDomain) {
-		return httpGet(chatty, listener, overrides, this.getApiURL(overrides) + "/swaggerapi/" + apiDomain + "/v1");
+		DefaultClient client = (DefaultClient) this.getClient(listener, getDisplayName(), overrides);
```
gabemontero: with my suggestion in `IOpenShiftPlugin` I believe this code path can remain unchanged

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -37,5 +58,41 @@ default ListBoxModel doFillWaitUnitItems() {
...
+                    sslCertificate(apiURL, auth.getCert()).
+                    build();
+
+            if (client == null) {
+                return FormValidation.error("Connection unsuccessful");
+            }
+
+            client.getServerReadyStatus();
+
+            IOpenShiftPlugin.httpGet(false, null, EnvVars.masterEnvVars, apiURL, authToken, "", client, auth, apiURL);
```
gabemontero: this will switch to `HttpUtils.httpGet`

gabemontero (commented):
Just a minor tweak and then I'll be GTM

gabemontero:
As an FYI, we'll wait until a future sprint / effort to see about further refactoring to move off of the static class to a lower level class/interface structure that just has `httpGet`.  @oatmealraisin and I discussed and came to this offline.

oatmealraisin:
@gabemontero I've added the class :) ptal?

src/main/java/com/openshift/jenkins/plugins/pipeline/model/HttpUtils.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+public class HttpUtils {
+    static String httpGet(boolean chatty, TaskListener listener, Map<String, String> overrides, String urlString, String authToken, String displayName, DefaultClient client, Auth auth, String apiURL) {
+        URL url = null;
+        try {
+            url = new URL(urlString);
+        } catch (MalformedURLException e) {
+            e.printStackTrace(listener.getLogger());
```
gabemontero: do a chatty check here as well

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -37,5 +58,41 @@ default ListBoxModel doFillWaitUnitItems() {
         return items;
     }
 
+    default FormValidation doTestConnection(@QueryParameter String apiURL, @QueryParameter String authToken) {
 
+
+        if (StringUtils.isEmpty(apiURL)) {
```
gabemontero, line 64: we should be able to fall back to seeing if "KUBERNETES_SERVICE_HOST" is set as an env var like the build steps do

gabemontero (commented):
Just a couple of minor nits

oatmealraisin:
@gabemontero I also made the key for the kube env var a final string in IOpenShiftPlugin, ptal?

gabemontero:
IGTM

A thought regarding testing .... can you find out, or do you already know, the URL of the HTTP request generated when you hit the test connection button?  If so, we should be able to add an extended test.

oatmealraisin:
@gabemontero Looking into it

PR #86, opened 2016-10-12
Modified code to check for whitespace

oatmealraisin:
Modified setters, contructors, and length checks to account for leading
and trailing whitespace.

@gabemontero @bparees ptal?

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/ImageStreamRevisionState.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/ImageStreamRevisionState.java
@@ -9,7 +9,7 @@ public class ImageStreamRevisionState extends SCMRevisionState implements Compar
 	public ImageStreamRevisionState(String commitId) {
 		if (commitId == null)
 			throw new IllegalArgumentException("nulls not allowed");
-		this.commitId = commitId;
+		this.commitId = commitId.trim();
 	}
 
 	@Override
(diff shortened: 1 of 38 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/TimedOpenShiftBasePostAction.java
```diff
@@ -12,8 +12,8 @@
 
     public TimedOpenShiftBasePostAction(String apiURL, String namespace, String authToken, String verbose, String waitTime, String waitUnit) {
         super(apiURL, namespace, authToken, verbose);
-        this.waitTime = waitTime;
-        this.waitUnit = waitUnit;
+        this.waitTime = waitTime.trim();
+        this.waitUnit = waitUnit.trim();
```
gabemontero: won't jobs from older version pass in null for the waitUnit?  I seem to recall that during our reviews there.
If my recollection holds , you can get a NullPointerException.

other existing params may that have been added over time may fall into that category as well.

Hence, if the above holds, I think you should do a global replace for each field, which will look like this example using `waitUnit`:

```
waitUnit != null ? waitUnit.trim() : null;
```

gabemontero (commented):
Just one comment, though if it holds, means a pervasive change for you.

Pending that, IGTM

oatmealraisin:
@gabemontero I've added the check throughout, good catch, thank you! ptal?

src/main/java/com/openshift/jenkins/plugins/pipeline/ImageStreamRevisionState.java
```diff
@@ -9,7 +9,7 @@
 	public ImageStreamRevisionState(String commitId) {
 		if (commitId == null)
 			throw new IllegalArgumentException("nulls not allowed");
-		this.commitId = commitId;
+		this.commitId = commitId != null ? commitId.trim() : null;
```
gabemontero: well, in this case, null is checked and exception thrown before hand in the line above ... we can just do commitId.trim() here.

gabemontero (commented):
Just 1 minor nit and I'm GTM.

The test job failed because of a vagrant hiccup.  We'll see if that hiccup is persistent after you push the update for the one nit.  If it persists I'll merge anyway and we can track the vagrant issue separately.

oatmealraisin:
@gabemontero I've fixed the unnecessary check, ptal?

gabemontero:
IGTM ... will wait for test job to hit button.

gabemontero:
no vargrant hiccup this time around ... job should complete soon

PR #87, opened 2016-10-12
Fixing default for destinationNamespace in tag step

jupierce:
Found issue in public version of plugin where tagger will default to the jenkins/OpenShift namespace if destinationNamespace is not supplied. This should fix it.

ptal @gabemontero @bparees

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
@@ -17,272 +17,269 @@ import java.util.*;
 
 public interface IOpenShiftImageTagger extends IOpenShiftPlugin {
 
-	final static String DISPLAY_NAME = "Tag OpenShift Image";
+    final static String DISPLAY_NAME = "Tag OpenShift Image";
 
-	default String getDisplayName() {
-		return DISPLAY_NAME;
-	}
-	
-	String getAlias();
-	
-	String getSrcTag();
-	
-	String getDestTag();
-		
-	String getSrcStream();
-	
-	String getDestStream();
-	
-	String getDestinationNamespace();
-		
-	String getDestinationAuthToken();
-		
-	default String getAlias(Map<String,String> overrides) {
-		return getOverride(getAlias(), overrides);
-	}
-	
-	default String getSrcTag(Map<String,String> overrides) {
-		return getOverride(getSrcTag(), overrides);
-	}
-	
-	default String getDestTag(Map<String,String> overrides) {
-		return getOverride(getDestTag(), overrides);
-	}
-	
-	default String getSrcStream(Map<String,String> overrides) {
-		return getOverride(getSrcStream(), overrides);
-	}
-	
-	default String getDestStream(Map<String,String> overrides) {
-		return getOverride(getDestStream(), overrides);
-	}
-	
-	default String getDestinationNamespace(Map<String,String> overrides) {
-		String val = getOverride(getDestinationNamespace(), overrides);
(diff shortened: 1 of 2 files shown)


jupierce:
Replaced whitespace with spaces, so advising PR review with w=1:
https://github.com/openshift/jenkins-plugin/pull/87/files?w=1

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
```diff
@@ -17,272 +17,274 @@
...
+    default String getSrcStream(Map<String, String> overrides) {
+        return getOverride(getSrcStream(), overrides);
+    }
+
+    default String getDestStream(Map<String, String> overrides) {
+        return getOverride(getDestStream(), overrides);
+    }
+
+    default String getDestinationNamespace(Map<String, String> overrides) {
+        return getOverride(getDestinationNamespace(), overrides);
```
bparees: doesn't the doc say this falls back to the source namespace if it's empty?

gabemontero: correct .... per the discussion @jupierce and I had on IRC, this needs to be
1) call getOverride(getDestin....), if non-empty return
2) return getNamespace(overrides)
getNamespace(overrides) should handle the fetching from NAMESPACE_ENV_VAR

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
```diff
@@ -17,272 +17,274 @@
...
+    String getDestinationAuthToken();
+
+    default String getAlias(Map<String, String> overrides) {
+        return getOverride(getAlias(), overrides);
+    }
+
+    default String getSrcTag(Map<String, String> overrides) {
+        String val = getOverride(getSrcTag(), overrides);
+        // If there is no source namespace, default to jenkins namespace
+        if (val.length() == 0 && overrides != null && overrides.containsKey(NAMESPACE_ENV_VAR)) {
```
gabemontero: I think this was meant for getDestinationNamespace .... that tag should not be derived from the namespace env var ... that said, we can just leverage getNamespace there

bparees:
also, in the future let's do large reformatting changes as their own commit.  this is going to be tough to look back at and understand the diff.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java
jupierce: It does, yes. This is handled by: https://github.com/openshift/jenkins-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java#L131

jupierce:
I think the confusion arises from where the defaulting logic taking place at https://github.com/openshift/jenkins-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftImageTagger.java#L131

The current PR does the following:
Source namespace order of precedence:
1. The overridden value for the sourceNamespace field
2. The value of the sourceNamespace field
3. The namespace hosting Jenkins

Destination namespace precedence:
1. The override value for the destinationNamespace field
2. The destinationNamespace field
3 The source namespace

gabemontero:
@jupierce and I looked over this at my cube.  He's removing the change he made to getSrcTag, and moving defaulting logic in the `coreLogic` method to `getDestinationNamespace`.

jupierce:
Removed erroneous defaulting logic from getSrcTag. Moved actual defaulting logic from coreLogic to getDestinationNamespace for clarity.

gabemontero:
IGTM ... will hit the button after the test job completes successfully

PR #88, opened 2016-10-13
Fail job when imagestream cannot be retrieved

csrwng:
- Interrupts the build and sets its result as Failed if there is no ImageStream found with the tag specified.
- Saves an ImageStreamRevisionState even when there is no commit-id. This is needed for the DSL step to continue polling (otherwise you have to manually start the build).
- Fixes the jelly file for the DSL step

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
@@ -9,10 +9,7 @@ import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.AbstractProject;
-import hudson.model.Job;
-import hudson.model.Run;
-import hudson.model.TaskListener;
+import hudson.model.*;
 import hudson.scm.*;
 import hudson.scm.PollingResult.Change;
 import hudson.util.FormValidation;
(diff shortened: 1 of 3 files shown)


csrwng:
@gabemontero ptal

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
```diff
@@ -130,7 +134,16 @@ public void checkout(Run<?, ?> build, Launcher launcher,
...
         if (chatty)
             listener.getLogger().println("\n\nOpenShiftImageStreams checkout called for " + bldName);
-        // nothing to actually check out in the classic SCM sense into the jenkins workspace
+
+        EnvVars env = build.getEnvironment(listener);
+        lastCommitId = getCommitId(listener, env);
+        if (lastCommitId == null) {
+            String imageStream = getImageStreamName(env);
+            String tag = getTag(env);
+            String msg = "Cannot find the last tagged image for ImageStream " + imageStream + " and tag " + tag;
```
gabemontero: The message should go into MessageConstants.java as a string with `%s` that we can then use String.format on.

gabemontero (commented):
Just one minor nit and then IGTM

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
csrwng: cool, will update

csrwng:
code updated

gabemontero:
IGTM

PR #94, opened 2016-10-18
Adding response object for openshiftExec DSL

### LINKED ISSUE #93: Is it possible to capture the outcome of oc exec
Hi all,

is it possible somehow to get the outcome of an oc exec in order to further decide how to proceed within the jenkins-pipeline.
Currently a call to openshiftExec does not seem to return anything - at the console I see what's happening, but is it possible to get the return status of the command which was executed?

Cheers

jupierce:
Fixes #93 

ptal @gabemontero @bparees @schoeffm

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,27 @@ Optional parameters are:
 -  "arguments" : Arguments for the command. May be specified as `arguments: [ 'arg1', 'arg2', ... ]` or `arguments: [ [ value: 'arg1' ], [ value : 'arg2' ], ... ]` 
 -  "waitTime" :  Time in milliseconds to wait for deployment completion.  Default is 3 minutes.
 
+The object returned exposes the following attributes:
+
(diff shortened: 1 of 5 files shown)


gabemontero:
IGTM @jupierce 

How do you want to handle coordination wrt your extended test?  At first blush I'm thinking merge what you have once it is fleshed out, where what you currently have does not inspect the return from the exec, and then create a separate pull for this change,  after I cut, toward the end of the week, the next version of the plugin with this change and other last minute fixes before dcut.

jupierce:
@gabemontero I like that approach. I definitely want to get the extended tests running on the existing cut before introducing this into the queue.

gabemontero:
ok cool - I'll defer cutting the new plugin release accordingly

schoeffm:
Great! Can't wait ... thx in advance!

PR #95, opened 2016-10-18
Changed test connection to check for null URL

oatmealraisin:
Small change for a null check

@gabemontero ptal?

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
@@ -1,6 +1,7 @@
 package com.openshift.jenkins.plugins.pipeline.model;
 
 
+import com.cloudbees.groovy.cps.Env;
 import com.openshift.internal.restclient.DefaultClient;
 import com.openshift.internal.restclient.authorization.AuthorizationContext;
 import com.openshift.internal.restclient.okhttp.OpenShiftAuthenticator;
@@ -11,6 +12,7 @@ import com.openshift.restclient.ClientBuilder;
 import com.openshift.restclient.http.IHttpConstants;
 import com.openshift.restclient.utils.SSLUtils;
 import hudson.EnvVars;
+import hudson.model.Computer;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 import okhttp3.Dispatcher;
@@ -61,13 +63,20 @@ public interface IOpenShiftPluginDescriptor {
     default FormValidation doTestConnection(@QueryParameter String apiURL, @QueryParameter String authToken) {
 
 
-        if (StringUtils.isEmpty(apiURL)) {
-            if(!EnvVars.masterEnvVars.containsKey(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY) &&
+        if (apiURL == null || StringUtils.isEmpty(apiURL)) {
+            if(EnvVars.masterEnvVars.containsKey(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY) &&
                     !StringUtils.isEmpty(EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY))) {
+                apiURL = EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY);
+
+                if(EnvVars.masterEnvVars.containsKey(IOpenShiftPlugin.KUBERNETES_SERVICE_PORT_ENV_KEY)) {
+                    apiURL = apiURL + ":" + EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_SERVICE_PORT_ENV_KEY);
+                }
+            } else if(EnvVars.masterEnvVars.containsKey(IOpenShiftPlugin.KUBERNETES_MASTER_ENV_KEY) &&
+                    !StringUtils.isEmpty(EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_MASTER_ENV_KEY))) {
+                apiURL = EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_MASTER_ENV_KEY);
+            } else {
                 return FormValidation.error("Required fields not provided");
             }
-
-            apiURL = EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY);
         }
 
         try {
@@ -91,9 +100,9 @@ public interface IOpenShiftPluginDescriptor {
 
             HttpUtils.httpGet(false, null, EnvVars.masterEnvVars, apiURL, authToken, "", client, auth, apiURL);
         } catch (MalformedURLException e ) {
-            return FormValidation.error("Connection Unsuccessful: Bad URL");
+            return FormValidation.error("Connection unsuccessful: Bad URL");
         } catch ( Exception e ) {
-            return FormValidation.error("Connection unsuccessful");
+            return FormValidation.error("Connection unsuccessful: " + e.getMessage());
         }
 
         return FormValidation.ok("Connection successful");
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuildVerifier.java
@@ -2,6 +2,7 @@ package com.openshift.jenkins.plugins.pipeline.dsl;
 
 import com.openshift.jenkins.plugins.pipeline.ParamVerify;
 import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftBuildVerifier;
+import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftPluginDescriptor;
 import hudson.Extension;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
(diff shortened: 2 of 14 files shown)


gabemontero:
IGTM

oatmealraisin:
@gabemontero EDIT: I noticed that the env vars set in my deploy for ephemeral didn't match the one we are checking.  I've updated that. Is this also relevant [here?](https://github.com/openshift/jenkins-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPlugin.java#L101-L109)

gabemontero:
Hmmm ... I see `KUBERNETES_SERVICE_HOST` referenced plenty doing a quick google search.  I think we need to better understand how you set up and why you don't see it but do see `KUBERNETES_MASTER`

How about you do a little research.  In this same setup you are testing with, follow the directions [here](https://github.com/openshift/origin/blob/master/examples/jenkins/README.md) and run the sample job with verbose set to true for one of the steps.  See if you can find either KUBERNETES_SERVICE_HOST or KUBERNETES_MASTER in the env var dump.  And if so, are their values the same or different.

oatmealraisin:
It looks like the jenkins templates set KUBERNETES_MASTER by default, [here](https://github.com/openshift/origin/blob/master/examples/jenkins/jenkins-persistent-template.json#L119-L122), perhaps this is incorrect?

gabemontero:
Ah, KUBERNETES_MASTER is one of the well known host names.  I _think_ I'd
rather check the service host env var before settling for this default, but
I want to better understand what is available when.

Per the test I suggested to you, I'm wondering if the Jenkins master env
var is not populated with all the env vars k8s dynamically seeded into the
pod at the time the DescriptorImpl is called when the test connection
button is hit.

I'll run the test as well.

oatmealraisin:
@gabemontero Using `oc env dc/jenkins --list`, it looks like KUBERNETES_SERVICE_HOST is not set in any of the deployment configs or pods during the process. 

We could do a staged check, checking KUBE_MASTER last?

gabemontero:
It is dynamically done.  I'll post an example when I get the chance
(possibly later tonight).  I'll run the sample I mentioned earlier.

gabemontero:
Here is a snippet from the sample job that has both env vars, along with a bunch of others (Note this is only a small subset of all the available env vars):

```
KUBERNETES_MASTER=https://kubernetes.default:443, KUBERNETES_PORT=tcp://172.30.0.1:443, KUBERNETES_PORT_443_TCP=tcp://172.30.0.1:443, KUBERNETES_PORT_443_TCP_ADDR=172.30.0.1, KUBERNETES_PORT_443_TCP_PORT=443, KUBERNETES_PORT_443_TCP_PROTO=tcp, KUBERNETES_PORT_53_TCP=tcp://172.30.0.1:53, KUBERNETES_PORT_53_TCP_ADDR=172.30.0.1, KUBERNETES_PORT_53_TCP_PORT=53, KUBERNETES_PORT_53_TCP_PROTO=tcp, KUBERNETES_PORT_53_UDP=udp://172.30.0.1:53, KUBERNETES_PORT_53_UDP_ADDR=172.30.0.1, KUBERNETES_PORT_53_UDP_PORT=53, KUBERNETES_PORT_53_UDP_PROTO=udp, KUBERNETES_SERVICE_HOST=172.30.0.1, KUBERNETES_SERVICE_PORT=443, KUBERNETES_SERVICE_PORT_DNS=53, KUBERNETES_SERVICE_PORT_DNS_TCP=53, KUBERNETES_SERVICE_PORT_HTTPS=443, KUBERNETES_TRUST_CERTIFICATES=true, KUBE_CA=/run/secrets/kubernetes.io/serviceaccount/ca.crt, KUBE_SA_DIR=/run/secrets/kubernetes.io/serviceaccount, LD_PRELOAD=libnss_wrapper.so, 
```

If not already there, can you add some trace that prints the entire set of env vars when in `httpGet`.

At this time, update your pull - I want us to check for KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT before falling back to KUBERNETES_MASTER.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -61,13 +61,13 @@ default ListBoxModel doFillWaitUnitItems() {
     default FormValidation doTestConnection(@QueryParameter String apiURL, @QueryParameter String authToken) {
 
 
-        if (StringUtils.isEmpty(apiURL)) {
-            if(!EnvVars.masterEnvVars.containsKey(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY) &&
-                    !StringUtils.isEmpty(EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY))) {
+        if (apiURL == null || StringUtils.isEmpty(apiURL)) {
+            if(!EnvVars.masterEnvVars.containsKey("KUBERNETES_MASTER") &&
```
gabemontero: So to record my review comment in line, still check for KUBERNETES_SERVICE_HOST first (and add KUBERNETES_SERVICE_PORT ), then if that does not exist, try KUBERNETES_MASTER

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -61,13 +61,13 @@ default ListBoxModel doFillWaitUnitItems() {
...
+                    !StringUtils.isEmpty(EnvVars.masterEnvVars.get("KUBERNETES_MASTER"))) {
```
gabemontero: Also, I'm wondering what you see when comparing `EnvVars.masterEnvVars` with `Computer.currentComputer().getEnvironment()` - perhaps my original suggestion to you was not ideal.

The typical job execution path seems to pull env vars from a number of sources.  Computer is one of them.

gabemontero:
@oatmealraisin - in doing some code searching / research for this PR, I noticed that on the DSL side of things, only com.openshift.jenkins.plugins.pipeline.dsl.OpenShiftBuilder.DescriptorImpl implements com.openshift.jenkins.plugins.pipeline.model.IOpenShiftPluginDescriptor.

Was that intentional?  I would think either all the DSL DescriptorImpl's implement that (to get the `doCheck...` and `doFillWaitUnits`) or none of them (since you won't get the "test connection" UI button in a pipeline script)

oatmealraisin:
@gabemontero Unfortunately I didn't check steps besides the builder, and assumed they behaved in the same way. I think that it's unintentional that the others do not implement the interface, so I'm adding the interface. I'm going to test the DSL steps a bit more to make sure that that's not the only bug. Good catch!

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
oatmealraisin: It looks like currentComputer() is null during the check logic

oatmealraisin:
@gabemontero I added a commit to patch the remaining DSL functionality. I also added a check for both env vars. Ptal?

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -61,13 +62,20 @@ default ListBoxModel doFillWaitUnitItems() {
...
 
 
-        if (StringUtils.isEmpty(apiURL)) {
-            if(!EnvVars.masterEnvVars.containsKey(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY) &&
+        if (apiURL == null || StringUtils.isEmpty(apiURL)) {
+            if(EnvVars.masterEnvVars.containsKey(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY) &&
                     !StringUtils.isEmpty(EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY))) {
+                apiURL = EnvVars.masterEnvVars.get(IOpenShiftPlugin.KUBERNETES_SERVICE_HOST_ENV_KEY);
+
+                if(EnvVars.masterEnvVars.containsKey("KUBERNETES_SERVICE_PORT")) {
```
gabemontero: create if it does not exist, otherwise use, a IOpenShiftPlugin constant like with service host

gabemontero: and bummer bout `currentComputer`

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -61,13 +62,20 @@ default ListBoxModel doFillWaitUnitItems() {
...
+                    apiURL = apiURL + ":" + EnvVars.masterEnvVars.get("KUBERNETES_SERVICE_PORT");
+                }
+            } else if(EnvVars.masterEnvVars.containsKey("KUBERNETES_MASTER") &&
```
gabemontero: another constant candidate

gabemontero (commented):
A couple of small nits, otherwise IGTM pending PR test validation

oatmealraisin:
@gabemontero Added the env vars, also added a message for when we catch Exceptions, so that we can better debug when they occur.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
oatmealraisin: Yeah I think it would have been more fine-grained if we could use that

gabemontero:
IGTM

PR #97, opened 2016-10-20
Prevent imagestream dsl step from continuing if imagestream not present

csrwng:
When using the imagestream dsl step, the jenkins job was not immediately aborting when the imagestream/tag could not be retrieved. (This was not the case with the non-dsl jobs)

This change makes it possible to abort the job while allowing the parent SCMStep to save the state information about the SCM in the job.

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
@@ -5,13 +5,22 @@ import com.openshift.jenkins.plugins.pipeline.model.IOpenShiftPluginDescriptor;
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.ResourceKind;
 import com.openshift.restclient.model.IImageStream;
+import hudson.AbortException;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.*;
-import hudson.scm.*;
+import hudson.model.AbstractProject;
+import hudson.model.Job;
+import hudson.model.Result;
+import hudson.model.Run;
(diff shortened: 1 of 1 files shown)


csrwng:
@gabemontero ptal

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
```diff
@@ -5,10 +5,7 @@
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.ResourceKind;
 import com.openshift.restclient.model.IImageStream;
-import hudson.EnvVars;
-import hudson.Extension;
-import hudson.FilePath;
-import hudson.Launcher;
+import hudson.*;
```
gabemontero: don't know if you did this or an IDE you are using, but I at least think explicitly listing each class is better than using the package wildcard.

thoughts?

gabemontero (commented):
one nit but otherwise IGTM

I'll wait until the current release cut completes, then if we've mitigated the nit comment, I'll merge this and start a new release cut.

src/main/java/com/openshift/jenkins/plugins/pipeline/OpenShiftImageStreams.java
csrwng: I agree. IntelliJ is trying to be too intelligent. Will update.

gabemontero:
btw the release publish process is done.  v1.0.24 is available.  I'll crank out the next version with this PR once we mitigate the nit comment.

csrwng:
man, that was quick, thx :)

gabemontero:
On Thursday, October 20, 2016, Cesar Wong notifications@github.com wrote:

> man, that was quick, thx :)

Likewise thx :-)

PR #100, opened 2016-10-28
Fix namespace setter for imagestream DSL step

csrwng:
Fixes the namespace setter for the ImageStream DSL step


csrwng:
@gabemontero ptal

PR #104, opened 2016-11-01
Modifying PR Testing Process

jupierce:
See openshift/origin#11711 for details. 
@bparees @gabemontero ptal

### DIFF
--- a/PR-Testing/README
+++ b/PR-Testing/README
@@ -1,17 +1,23 @@
 These files are leveraged by the test-openshift-pipeline-plugin job at https://ci.openshift.redhat.com/jenkins.
 A test image based off of the jenkins image at https://github.com/openshift/jenkins will be built with the dockerfile
-here and any openshift-pipeline.hpi file stored in the PR-Testing/jpi file as "openshift-pipeline.jpi".  In the case of
+here and any OpenShift Jenkins plugin file stored in the PR-Testing/jpi file as "openshift-pipeline.jpi".  In the case of
 the test-openshift-pipeline-plugin job, it will be the hpi file from your PR.  But you can also test any local changes
 you have to the plugin in the same way.
 
-The `docker build` command used, if run from the same directory as this README, and if an "openshift-pipeline.jpi" file
-exists in the "jpi" subdirectory, is:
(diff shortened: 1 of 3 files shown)


gabemontero:
IGTM ... I'll wait until the origin PR merges before hitting the button.

Thanks!!

bparees:
do we need to/should we update the PR test job to actually invoke this new image build script instead of whatever it is doing today?

gabemontero:
We'll minimally need to change the job to set `USE_SNAPSHOT_JENKINS_IMAGE=1`

Eyeballing the new script's contents, it isn't a necessity to call it from the PR test job.  The pros in doing so would be to validate the script.  The con would be that  PR could be submitted which changes the script in such a way as to mess up the testing.

The con has more weight with me than the pro in this case.  But certainly we can discuss more if desired.

bparees:
> The con would be that PR could be submitted which changes the script in such a way as to mess up the testing.

i think that's a pro.  having the PR tester actually testing the image build script seems useful.

if your concern is that someone breaks the image build script in such a way that:
1) it still succeeds (so the PR test doesn't fail because it couldn't build the image)
2) it doesn't actually build the image

which results in the PR test not actually testing a newly built image, that's fair, but i think we should just enforcing that a new image gets built successfully as part of the PR testing.  I don't like the idea of having an image builder script, but not using it and instead re-implementing that same logic directly in the CI job definition.

gabemontero:
You captured my concern.  I'm afraid the newly built plugin won't get tested because the PR messes up the script that ensures it is tested.

The _new_ script provided feels more like helper to me than something "official".  Admittedly subjective distinctions.

I'll look at / consider both approaches when the origin PR merges and we complete the loop here.

jupierce:
Based on that feedback, I believe I can resolve the concern by making the extended tests blow IF USE_SNAPSHOT_JENKINS_IMAGE=1 but no snapshot image is available.

jupierce:
@gabemontero blow = The extended tests will fail if USE_SNAPSHOT_JENKINS_IMAGE is defined but there is no jenkins snapshot image detected. That should detect the scenario where the build script is crippled somehow.

gabemontero:
@jupierce ok cool thx for the clarification

PR #105, opened 2016-11-01
Forcing error in Jenkins builds on exec error/fail

jupierce:
QE requested to fail job if there are errors in the underlying exec API: https://bugzilla.redhat.com/show_bug.cgi?id=1390491

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public interface IOpenShiftExec extends ITimedOpenShiftPlugin, IPodExec.IPodExecOutputListener {
 
@@ -82,6 +83,8 @@ public interface IOpenShiftExec extends ITimedOpenShiftPlugin, IPodExec.IPodExec
 
         final CountDownLatch latch = new CountDownLatch(1);
 
+        final AtomicBoolean reportError = new AtomicBoolean(false);
+
         IStoppable exec = p.accept(new CapabilityVisitor<IPodExec, IStoppable>() {
             @Override
             public IStoppable visit(IPodExec capability) {
@@ -107,6 +110,7 @@ public interface IOpenShiftExec extends ITimedOpenShiftPlugin, IPodExec.IPodExec
 
                     @Override
                     public void onExecErr(String message) {
+                        reportError.set(true);
                         listener.error("Error during exec: " + message);
                         IOpenShiftExec.this.onExecErr(message);
                     }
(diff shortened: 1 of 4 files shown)


jupierce:
ptal @gabemontero

jupierce:
The new behavior should be as follows:
- Timeouts/failures/errors should cause traditional builds to fail.
- DSL jobs will only fail because of exec timeouts. This should allow DSL scripts to handle errors via scripting.

bparees:
lgtm but will definitely defer to @gabemontero

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
```diff
@@ -130,19 +135,25 @@ public void onFailure(IOException e) {
...
         try {
-            success = latch.await(Math.max(remainingWaitTime, 1), TimeUnit.MILLISECONDS);
+            timeout = ! latch.await(Math.max(remainingWaitTime, 1), TimeUnit.MILLISECONDS);
         } catch (InterruptedException ie) {
         }
 
-        if ( success ) {
-            listener.getLogger().println(String.format(MessageConstants.EXIT_EXEC_GOOD, DISPLAY_NAME));
+        if ( timeout ) {
+            listener.getLogger().println(String.format(MessageConstants.EXIT_EXEC_BAD, DISPLAY_NAME, "Timeout waiting for exec to finish"  ));
```
gabemontero: create a message constant for "timeout waiting for exec to finish"

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
```diff
@@ -130,19 +135,25 @@ public void onFailure(IOException e) {
...
+            return false;
+        }
+
+        if ( reportError.get() ) {
+            listener.getLogger().println(String.format(MessageConstants.EXIT_EXEC_BAD, DISPLAY_NAME, "Error or failure during exec"));
```
gabemontero: create a message constant for "error or failure during exec"

gabemontero (commented):
Couple of minor comments.

Also, the README should be updated to reflect the behavior here, especially given the diff between freestyle and pipeline.

Otherwise, LGTM.  I'm amenable to merging this evening and letting @jupierce handle the comments in a subsequent PR tomorrow if we are anxious to let QE validate the change.  Just let me know - thanks.

bparees:
> I'm amenable to merging this evening and letting @jupierce handle the comments in a subsequent PR tomorrow if we are anxious to let QE validate the change. Just let me know - thanks.

i think we're fine to wait.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
jupierce: Will do.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
jupierce: Will do.

jupierce:
Additions to README.MD and new string constants in last push. 

Types of errors:
(1) Pod not present
(2) Specified container not present in pod
(3) Command not found  (e.g. "echo1" vs "echo")
(4) Timeout

Freestyle job killed by: 1, 2, 3, 4
DSL job killed by: 1, 5

gabemontero:
IGTM - even though @jupierce 's extended test fixes got in before this started, I need to make the tweaks to the PR test job for everything to take hold.  So not waiting on this PR test job.

PR #113, opened 2016-11-11
Make use of resource watch instead of Thread.sleep

jcantrill:
This PR removes the use of Thread.sleep in a few places in lieu of watchers that do OpenShift watches

cc @gabemontero any chance you can help me verify the changes work.  I was able to get logs but my origin code is unable to see github.com in order to confirm its actually able to properly monitor a build.


gabemontero:
yep taking a look / can do some testing @jcantrill - as we discussed in IRC, will probably merge / cut a new release after 3.4 code freeze

gabemontero:
@jupierce FYI some suggestions from @jcantrill

gabemontero:
@jcantrill the change has an couple of initial functional issues:
1) it always dumps the build log, even if the build step has disabled that
2) there is a lot of duplication of log output ... given lines being printed multiple times ... both @jupierce and myself have had to do some work to avoid this from occurring

If you want to debug those issues, great.  If not, we can certainly use this PR as a starting point to leveraging watches in upcoming versions of the plugin

just let us know ... and thanks a bunch for putting this together

jcantrill:
I am willing to debug... i wonder if this is because of the code that calls 'waitFor' which is in a loop.  I started modifying that but stopped.  Can you point me how you tested?  Create a BC and then just trigger a build from jenkins?  Monitor log output, etc?

PR #118, opened 2016-11-21
allow start of pipeline strategy builds

### LINKED ISSUE #117: openshiftBuild step does not recognize when a pipeline build has completed
Use `openshiftBuild` to start a new build with a pipeline strategy. 

Expected:
Once the pipeline build finishes, the openshiftBuild step completes

Actual:
The openshiftBuild step never recognizes that the pipeline build has finished and times out.

gabemontero:
fixes https://github.com/openshift/jenkins-plugin/issues/117

@csrwng @bparees FYI

(note, on the file diff, another case where `?w=1` on the file diff URL looks to be useful)

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
@@ -109,25 +109,18 @@ public interface IOpenShiftBuilder extends ITimedOpenShiftPlugin {
         }
         return bld;
     }
-
-    default void waitOnBuild(IClient client, long startTime, String bldId, TaskListener listener, Map<String, String> overrides, long wait, boolean follow, boolean chatty, IPod pod) throws InterruptedException {
-        IBuild bld = null;
-        String bldState = null;
-
-        // get internal OS Java REST Client error if access pod logs while bld is in Pending state
-        // instead of Running, Complete, or Failed
-
+    
+    default IStoppable getBuildPodLogs(IClient client, String bldId, Map<String, String> overrides, boolean chatty, TaskListener listener, AtomicBoolean needToFollow) {
         IStoppable stop = null;
+        List<IPod> pods = client.list(ResourceKind.POD, getNamespace(overrides));
+        for (IPod pod : pods) {
+            if (chatty)
+                listener.getLogger().println("\nOpenShiftBuilder found pod " + pod.getName());
 
-        final AtomicBoolean needToFollow = new AtomicBoolean(follow);
-
-        while (System.currentTimeMillis() < (startTime + wait)) {
-            bld = client.get(ResourceKind.BUILD, bldId, getNamespace(overrides));
-            bldState = bld.getStatus();
-            if (Boolean.parseBoolean(getVerbose(overrides)))
-                listener.getLogger().println("\nOpenShiftBuilder bld state:  " + bldState);
-
-            if (needToFollow.compareAndSet(true, false)) {
+            // build pod starts with build id
+            if (pod.getName().startsWith(bldId)) {
+                if (chatty)
+                    listener.getLogger().println("\nOpenShiftBuilder going with build pod " + pod);
                 final String container = pod.getContainers().iterator().next().getName();
 
                 stop = pod.accept(new CapabilityVisitor<IPodLogRetrievalAsync, IStoppable>() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
```diff
@@ -128,35 +128,46 @@ default void waitOnBuild(IClient client, long startTime, String bldId, TaskListe
...
-                                // If follow fails, try to restart it on the next loop.
-                                needToFollow.compareAndSet(false, true);
-                            }
-                        }, new IPodLogRetrievalAsync.Options()
-                                .follow()
-                                .container(container));
-                    }
-                }, null);
+              List<IPod> pods = client.list(ResourceKind.POD, getNamespace(overrides));
+              for (IPod pod : pods) {
```
bparees: if you get here before the pod is created (for normal build scenarios), aren't you going to skip over showing the build logs entirely?  (the previously logic looped waiting for a pod to be created, that doesn't seem to happen now?)

gabemontero: We don't break out the while loop unless the build is completed.  In my testing, I did not hit a scenario where for a non-jenkinsfile strategy, the build completed and the build pod did *NOT* exist.

Hence, the build logs showed up.

As a side bar, if I attempted to show logs for a jenkinsfile strategy, it did not fail.  But of course there were no logs to dump.

bparees: hmm, i missed that, but it still seems like there's a timing window in which there are no pods showing up, then the build shows as complete and the code breaks out, w/o having noticed the build pod and showing its logs.

I wonder if it would be better if the code actually understood whether this is a build that should produce a pod or not, and handled it accordingly?

gabemontero: On whether the BC should produce a Pod, I did look into that ...the IBuildConfig surfaced by the restclient does not surface the necessary information in that regard.  To me at least, it was low enough priority to not deal with it at this time.  Among other things, I'd rather not pull in a SNAPSHOT version of the restclient prior to 3.4 shipping if I can help it.

Let me know how that affects your opinion.

I'll tinker with your timing window in the interim.

bparees: it doesn't provide the strategy for the buildconfig?

gabemontero: That is what I was saying ... but I just double-checked again and now see it :-)

I'll update to leverage that.

gabemontero: Well, it is a little kludgy ... currently it is only set for source/docker/custom.  If "jenkinsfile" it is null.  I'll work with that, and add a TODO to update the restclient down the road.

gabemontero: updates pushed

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
```diff
@@ -158,9 +151,36 @@ public void onFailure(IOException e) {
...
+        final AtomicBoolean needToFollow = new AtomicBoolean(follow);
+
+        while (System.currentTimeMillis() < (startTime + wait)) {
+            bld = client.get(ResourceKind.BUILD, bldId, getNamespace(overrides));
+            bldState = bld.getStatus();
+            if (Boolean.parseBoolean(getVerbose(overrides)))
+                listener.getLogger().println("\nOpenShiftBuilder bld state:  " + bldState);
+
+            if (needToFollow.compareAndSet(true, false)) {
+                stop = this.getBuildPodLogs(client, bldId, overrides, chatty, listener, needToFollow);
```
bparees, line 177: isn't calling this multiple times going to result in duplicated log output?

gabemontero: Nope, the second call only occurs if the first did not find the pod (the `stop == null` check)

gabemontero:
The test job https://ci.openshift.redhat.com/jenkins/job/test-openshift-pipeline-plugin/119/consoleFull actually passed:
```
Ran 11 of 573 Specs in 3663.079 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 562 Skipped Nov 21 19:18:57.552: INFO: Error running cluster/log-dump.sh: fork/exec /data/src/github.com/openshift/origin/vendor/k8s.io
```

But the job is reporting a failure for a reason I am not entirely gleaming yet.  I'll solicit some help on a separate thread for that piece.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
bparees: I'm referring to the fact that that you go through this loop multiple times as you wait for the build to finish.

gabemontero:
Ok the same objects are returned on the multiple calls. We aren't getting
duplicate logging. But yes i can see about prefacing both calls with if
stop == null and confirm the behavior. Assuming it works ok i will push
that change.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
gabemontero: Added debug to confirm (both a print prior to `this.getBuildPodLogs` in line 176 and a temporary print in a temporary else for the if in line 175) ... the AtomicBoolean `needToFollow.compareAndSet(true, false)` ensures that the `this.getBuildPodLogs(...` on line 176 only gets called once in the loop as we go through multiple times.

gabemontero:
I did push an update, adding a break in the new getBuildPodLogs method after finding the pod.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftBuilder.java
bparees: ah, it's the compareAndSet bit that I was missing.

csrwng:
LGTM

gabemontero:
thanks gentlemen

gabemontero:
@csrwng - if you like, there is a built plugin with this change at https://ci.openshift.redhat.com/jenkins/job/openshift-pipeline-plugin/33/s3/download/openshift-pipeline.hpi

@bparees - would you like to have a new version of the plugin cut for this change now, or hold (perhaps until post 3.4) ?

bparees:
> @bparees - would you like to have a new version of the plugin cut for this change now, or hold (perhaps until post 3.4) ?

up to you and how risky you feel it is.  we have runway to get it in.

gabemontero:
Feels low risk. I'll get started on cutting a release when I get back.

Issue #99, opened 2016-10-26
document release process


bparees:
@gabemontero can you add a RELEASING.md doc that covers the steps required to build and publish a new version of the plugin to jenkinsci?

Does not have to cover anything about updating the jenkins image itself, just what it takes to go from merging a commit to releasing a new plugin version.

I realize not everyone will have permissions to actually do so, but at least we'll know the steps and can grant people permission as needed.

PR #119, opened 2016-11-22
document release process

gabemontero:
Fixes https://github.com/openshift/jenkins-plugin/issues/99

@bparees FYI

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,32 @@
+# Process for cutting release of this plugin
+
+As noted in this repository's README.md, the documentation and code at [https://github.com/openshift/jenkins-plugin](https://github.com/openshift/jenkins-plugin) always hosts the very latest version, including possibly pre-released versions that are still under test.
+The associated repository under the JenkinsCI project, [https://github.com/jenkinsci/openshift-pipeline-plugin](https://github.com/jenkinsci/openshift-pipeline-plugin), is only updated as part of cutting 
+official releases of this plugin.
+
+To cut a new release of this plugin, first perform a `git clone` of [https://github.com/jenkinsci/openshift-pipeline-plugin](https://github.com/jenkinsci/openshift-pipeline-plugin), 
+
+## Set up local repository to cut release:
+
+1. From the parent directory you've chosen for you local repository, run `git clone git@github.com:jenkinsci/openshift-pipeline-plugin.git`
+1. Change directories into `openshift-pipeline-plugin`, and run `git remote add upstream git://github.com/openshift/jenkins-plugin`
+1. Then pull the latest changes from [https://github.com/openshift/jenkins-plugin](https://github.com/openshift/jenkins-plugin) with the following:
+
+	$ git checkout master
+	$ git fetch upstream
+	$ git fetch upstream --tags
+	$ git rebase upstream/master
+	$ git push origin master
+	$ git push origin --tags
+
+
+## Submit the new release to the Jenkins organization
+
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -0,0 +1,31 @@
...
+git rebase upstream/master
+git push origin master
+git push origin --tags
+```
+
+## Submit the new release to the Jenkins organization
+
+1. Assumptions: your Git ID has push access to the two repositories for this plugin; your Jenkins ID (https://wiki.jenkins-ci.org/display/JENKINS/User+Account+on+Jenkins) is listed in https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/plugin-openshift-pipeline.yml
+1. Then run `mvn release:prepare release:perform`
+	- NOTE:  You'll minimally be prompted for the `release version`, `release tag`, and the `new development version` (typically a version indicator that ends with `-SNAPSHOT`).  Also, if we happen to need to depend on a SNAPSHOT version of openshift-restclient-java, you'll get prompted about moving off the SNAPSHOT version, but override this.  Stay on the SNAPSHOT version of openshift-restclient-java if recent updates to that dependency that we need have not yet been cut to a new FINAL release.
```
bparees: the instructions about snapshot vs final, when/if you'll get prompted, and what you should/shouldn't do if you do get prompted, are a bit confusing.

also some guidance about what values to use for release version, release tag, and new development version would be helpful.

RELEASING.md
```diff
@@ -0,0 +1,31 @@
...
+1. Run `git push https://github.com/openshift/jenkins-plugin.git master` to upload the 2 commits created for cutting the new release to our upstream, development repository, and get the two repositories back in sync.
+1. Monitor https://updates.jenkins-ci.org/download/plugins/openshift-pipeline/ for the existance of the new version of the plugin.  Warning: the link for the new version will show up, but does not mean the `openshift-pipeline.hpi` file is available yet.  Click the link to confirm you can download the new version of the plugin.  When you can download the latest `openshift-pipeline.hpi` file, the process is complete, and the new release is available.
```
bparees: s/existance/existence/

gabemontero: just pushed correction

RELEASING.md
gabemontero: OK I've attempted to clarify / clear up.  Let me know if it makes better sense to humans besides me :-)

RELEASING.md
```diff
@@ -0,0 +1,31 @@
...
+1. Then run `mvn release:prepare release:perform`
+	- NOTE:  You'll minimally be prompted for the `release version`, `release tag`, and the `new development version`.  Default choices will be provided for each, and the defaults are typically acceptable, so you can just hit the enter key for all three prompts.  As an example, if we are currently at v1.0.36, it will provide 1.0.37 for the new `release version` and `release tag`.  For the `new development version` it will provide 1.0.38-SNAPSHOT, which is again acceptable.  The only time you *might* have to override the defaul provided is if we happen to need to depend on a SNAPSHOT version of openshift-restclient-java (i.e. `5.3.0-SNAPSHOT`).  This occurs when we add new features to openshift-restclient-java, but the eclipse team has not cut a new, official release (which will typically look like `5.3.0-FINAL`).  If we are in such a mode, you'll get prompted about moving off the SNAPSHOT version (the default provided would be `5.3.0`), but override this (i.e. type in `5.3.0-SNAPSHOT`).
```
bparees: s/defaul/default/

bparees: s/i.e./e.g./  ("that is to say" vs "for example")

bparees: s/if we happen to need to depend on/if we currently depend on/

bparees: (the i.e.->e.g. comment only applies to the first instance, the second usage is correct, or at least can be correct either way in my head)

bparees:
couple more suggestions and lgtm.

RELEASING.md
gabemontero: Updates incorporated ... about to push

gabemontero:
Updates pushed ... also some markdown lang cleanup based on the markdown preview tab I found in eclipse

PR #120, opened 2016-12-08
how to query service nodeport in jenkins?

zhangzaizai:
In jenkins, I am create a mysql deploymentconfig and create service with type LoadBalancer and nodeport:0, but i want access mysql with route name and nodeport, but jenkins openshift plugin create resource has no return, so, how can i query a service nodeport with openshift plugins ?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
   </parent>
   <groupId>com.openshift.jenkins</groupId>
   <artifactId>openshift-pipeline</artifactId>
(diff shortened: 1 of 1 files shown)


gabemontero:
@zhangzaizai - all you have pushed are updates to the version in the pom.xml.  

You need to push any actual code or doc changes you have made.

Also, we don't update the pom.xml version items in this repo.  I'll do that as part of updating our jenkinsci fork as part of cutting a new release.

PR #123, opened 2017-01-18
provide some form of ssl hostname validation

gabemontero:
@bparees @csrwng PTAL

This stems from the hornets' nest I stirred up in https://github.com/openshift/origin/issues/12476 with comment https://github.com/openshift/origin/issues/12476#issuecomment-272461025

Clayton ultimately chimed in with comment https://github.com/openshift/origin/issues/12476#issuecomment-272547257, asking we do something asap.

Based on my parsing of some SSL related javadoc, and analysis of results from the base `oc cluster up` start up, I've crafted this change to provide some hostname based validation of the remote/server side of the SSL connection.  

Aside from using the hostname that the plugin uses for the api server, trying to interpret some proxy based comments I thought Ben made somewhere, I also added the ability to manually configure a list of hosts that the remote certs needed to ref.

Thoughts?

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
@@ -16,6 +16,7 @@ import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.logging.Logger;
 
 import javax.net.ssl.SSLSession;
 import javax.net.ssl.TrustManager;
@@ -24,13 +25,12 @@ import javax.net.ssl.X509TrustManager;
 
 import com.openshift.restclient.ISSLCertificateCallback;
 
-import hudson.EnvVars;
-import hudson.model.AbstractBuild;
-import hudson.model.Run;
 import hudson.model.TaskListener;
 
 
 public class Auth implements ISSLCertificateCallback {
+    static final Logger LOGGER = Logger.getLogger(Auth.class.getName());
+    
 	private static final String AUTH_FILE = "/run/secrets/kubernetes.io/serviceaccount/token";
 	private static final String CERT_FILE = "/run/secrets/kubernetes.io/serviceaccount/ca.crt";
 	public static final String CERT_ARG = " --certificate-authority=/run/secrets/kubernetes.io/serviceaccount/ca.crt ";
@@ -70,9 +70,7 @@ public class Auth implements ISSLCertificateCallback {
 		return auth;
 	}
 	
-	// let's only use the local trust store within a calling method, so callers will sync on Auth, get the store
-	// and then reset ... assumption that the sync cost impact is tolerable
-	
+	// we use a local trust store when we have to make direct http get's and bypass the openshift-restclient-java
 	public static X509TrustManager createLocalTrustStore(Auth auth, String apiURL) {
 		if (auth.useCert()) {
 			try {
@@ -110,7 +108,6 @@ public class Auth implements ISSLCertificateCallback {
 		x509TrustManager = null;
 	}
 	
-	@Override
 	public boolean allowCertificate(final X509Certificate[] certificateChain) {
 		// this will be called if the trustManager.checkServerTrusted call fails in openshift-restclient-java;
 		// we are in this path for the "handle untrusted certifcates" or "skip tls verify" path
@@ -125,6 +122,7 @@ public class Auth implements ISSLCertificateCallback {
 			return true;
 		}
 
+	    // we use a local trust store when we have to make direct http get's and bypass the openshift-restclient-java
 		if (x509TrustManager != null) {
 			try {
 				x509TrustManager.checkServerTrusted(certificateChain, "RSA");
@@ -146,12 +144,22 @@ public class Auth implements ISSLCertificateCallback {
 	}
 	
 
-	@Override
 	public boolean allowHostname(String hostname, SSLSession sslSession) {
-		//mimic SSLCertificateCallback implementation from jbosstools-openshift repo - we can noop
-		//also lines up with what was observed in k8s jennkins plugin
-		return true;
+	    // while the old openshift-restclient ISSLCertificateCallback method still has an allowHostname,
+	    // we added support there so that we could seed the underlying okhttp3 code to use its default
+	    // hostname verifier; as such, this method should only be called if we are in skip tls mode, in 
+	    // which case we will bypass the okhttp3 hostname verification (it does not have a skip tls options)
+	    if (listener != null)
+	        listener.getLogger().println("Auth - allowHostname getting called, skip tls is " + skipTls);
+	    if (skipTls)
+	        return true;
+	    else {
+	        // should not be getting called if skip tls is false ... err on side of safety, refuse the hostname
+            LOGGER.info("allowHostname getting called but we are not skipping tls; returning false; investigate openshift-restclient-java and okhttp3");
+            return false;
+	    }
 	}
+	
 	public boolean useCert() {
 		return cert != null && !skipTls;
 	}
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,8 @@
 	<dependency>
 	    <groupId>com.openshift</groupId>
 	    <artifactId>openshift-restclient-java</artifactId>
-	    <version>5.2.0.Final</version>
+	    <!--  <version>5.4.0.Final</version>-->
+	    <version>5.4.0-SNAPSHOT</version>
 	</dependency> 
     
   	<dependency>
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPlugin.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPlugin.java
@@ -136,25 +136,20 @@ public interface IOpenShiftPlugin extends IOpenShiftParameterOverrides {
     }
 
     boolean coreLogic(Launcher launcher, TaskListener listener, Map<String, String> overrides) throws InterruptedException;
-
+    
     default IClient getClient(TaskListener listener, String displayName, Map<String, String> overrides) {
-        IClient client = new ClientBuilder(getApiURL(overrides)).
-                sslCertificateCallback(getAuth()).
-                usingToken(Auth.deriveBearerToken(getAuthToken(overrides), listener, Boolean.getBoolean(getVerbose(overrides)), overrides)).
-                sslCertificate(getApiURL(overrides), getAuth().getCert()).
-                build();
-        if (client == null) {
-            listener.getLogger().println(String.format(MessageConstants.CANNOT_GET_CLIENT, displayName, getApiURL(overrides)));
-        }
-        return client;
+        return this.getClient(listener, displayName, overrides, Auth.deriveBearerToken(getAuthToken(overrides), listener, Boolean.getBoolean(getVerbose(overrides)), overrides));
     }
 
     default IClient getClient(TaskListener listener, String displayName, Map<String, String> overrides, String token) {
-        IClient client = new ClientBuilder(getApiURL(overrides)).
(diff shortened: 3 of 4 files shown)


bparees:
we should probably get @liggitt or @deads2k to take a look.  Superficially it looks ok to me, but I really don't know this space.  In particular the proxy host thing may not be the right way to handle the potential interference of proxies (it's probably assumed any proxy would be a pass-through proxy, so no special handling is needed) but it does provide a nice "get out of jail free" way to bypass cert verification for specific hostnames.

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
```diff
@@ -148,9 +157,47 @@ public boolean allowCertificate(final X509Certificate[] certificateChain) {
...
+            }
+        }
+        
+	    try {
+            for (String h : lookFor) {
+                // the hostname passed here is the hostname we use to initiate the ssl flow;
+                // note the javadoc for SSLSession.getPeerHost() says "This value is not authenticated and should not be relied upon";
+                // note SSLSession.getPeerPrincipal.getName() based on testing returns a CN=... subject name not useful to compare with;
+                // so we see if our hostname is in the list of any of ssl session's cert's subject names ... if so, we bless this
+                Certificate[] certs = sslSession.getPeerCertificates();
```
liggitt: only the first certificate should be checked, since it's the only one the server has proven it holds a private key for. Any additional certificates are ones the server sent, but does not hold a private key for (they're to be used as intermediates when testing a valid signing chain from the first cert to a trusted root cert)

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
```diff
@@ -148,9 +157,47 @@ public boolean allowCertificate(final X509Certificate[] certificateChain) {
...
+                for (Certificate cert : certs) {
+                    if (cert instanceof X509Certificate) {
+                        X509Certificate x509cert = (X509Certificate)cert;
+                        Collection<List<?>> subjNames = x509cert.getSubjectAlternativeNames();
```
liggitt: you need to include the CN component of the cert subject

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
```diff
@@ -11,38 +11,47 @@
...
 import hudson.model.TaskListener;
 
 
 public class Auth implements ISSLCertificateCallback {
+    static final Logger LOGGER = Logger.getLogger(Auth.class.getName());
+    
 	private static final String AUTH_FILE = "/run/secrets/kubernetes.io/serviceaccount/token";
 	private static final String CERT_FILE = "/run/secrets/kubernetes.io/serviceaccount/ca.crt";
 	public static final String CERT_ARG = " --certificate-authority=/run/secrets/kubernetes.io/serviceaccount/ca.crt ";
+	public static final String SSL_PROXY_HOSTS = "SSL_PROXY_HOSTS";
```
liggitt: is this a standard thing?

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
```diff
@@ -148,9 +157,47 @@ public boolean allowCertificate(final X509Certificate[] certificateChain) {
 
 	@Override
 	public boolean allowHostname(String hostname, SSLSession sslSession) {
```
liggitt, line 147: I can't tell if this is called for every connection, or just ones whose certs which fail standard hostname verification? is this implementing https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/HostnameVerifier.html?

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
```diff
@@ -148,9 +157,47 @@ public boolean allowCertificate(final X509Certificate[] certificateChain) {
...
+                // note SSLSession.getPeerPrincipal.getName() based on testing returns a CN=... subject name not useful to compare with;
+                // so we see if our hostname is in the list of any of ssl session's cert's subject names ... if so, we bless this
+                Certificate[] certs = sslSession.getPeerCertificates();
+                for (Certificate cert : certs) {
+                    if (cert instanceof X509Certificate) {
+                        X509Certificate x509cert = (X509Certificate)cert;
+                        Collection<List<?>> subjNames = x509cert.getSubjectAlternativeNames();
+                        if (subjNames != null) {
+                            for (List<?> list : subjNames) {
+                                if (list.contains(h)) {
```
liggitt: normal matching rules are described in https://tools.ietf.org/html/rfc6125#section-6.4

the biggest differences from here are case-insensitivity and wildcard matching for DNS SANs or CNs. For example, a cert with a DNS SAN of `*.foo.com` should allow a hostname of `www.foo.com`

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
```diff
@@ -148,9 +157,47 @@ public boolean allowCertificate(final X509Certificate[] certificateChain) {
...
+                // note the javadoc for SSLSession.getPeerHost() says "This value is not authenticated and should not be relied upon";
+                // note SSLSession.getPeerPrincipal.getName() based on testing returns a CN=... subject name not useful to compare with;
+                // so we see if our hostname is in the list of any of ssl session's cert's subject names ... if so, we bless this
+                Certificate[] certs = sslSession.getPeerCertificates();
+                for (Certificate cert : certs) {
+                    if (cert instanceof X509Certificate) {
+                        X509Certificate x509cert = (X509Certificate)cert;
+                        Collection<List<?>> subjNames = x509cert.getSubjectAlternativeNames();
+                        if (subjNames != null) {
+                            for (List<?> list : subjNames) {
```
liggitt: this list isn't what you think it is... it contains two items, the first of which is an Integer telling you what type of subjectAlternativeName this is. You should only check DNS (type 2) and IP Address (type 7). The second element is the data you should compare the host against.

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
gabemontero: No, this is just a backdoor config to provide a specific list of hosts to verify against if the solution overlooked some valid scenario.

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
gabemontero: It is an implementation of https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/HostnameVerifier.html

A potentially interesting point on *when* this is called.  For ease, I'll paste a comment / piece of analysis from the referenced issue driving this:

Per the javadoc at https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/HostnameVerifier.html, there are no default implementations in the JDK.  I also see similar discussions in http://stackoverflow.com/questions/6031258/java-ssl-how-to-disable-hostname-verification .. ("There is no hostname verification in standard Java SSL sockets or indeed SSL, so that's why you can't set it at that level. Hostname verification is part of HTTPS (RFC 2818): that's why it manifests itself as javax.net.ssl.HostnameVerifier, which is applied to an HttpsURLConnection.")

There seems to be an inconsistency between what the `HostnameVerifier` javadoc says (dating back to java 1.4), spefically:

`During handshaking, if the URL's hostname and the server's identification hostname mismatch, the verification mechanism can call back to implementers of this interface to determine if this connection should be allowed. `

But what I see in the lower level http libraries that openshift-restclient and fabric8 use (okhttp3) does not seem to adhere to that.  If you look at https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java#L266-L277, it calls verify irregardless of the outcome of the handshake.

So it would seem this will get called on every connection, unless the base handshake *fails*, in which case the exception is thrown and it bypasses the call to the verify method.

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
gabemontero: Yep I noticed the 2 items when prototyping this, but was unaware of the significance of type 2 / type 7.  I'll put in additional checks along those lines.

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
gabemontero: Gotcha - thanks.  Will work on updates.

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
gabemontero: 10-4 ... will be part of next update.

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
gabemontero: Apologies, I'm not following.  If I do a toString() on the cert, I see this:
```
Subject: CN=10.13.137.191
```
It is the only place where `CN` is present.

Are you saying I need to parse out the `10.13.137.191` piece from the Subject and validate it as the address of the api master?

liggitt: yes :(

gabemontero:
Going to do an about face on this one.  In wrapping up the changes to @liggitt 's comments, I discovered that okhttp3 does in fact have a default hostname verifier !!!

And it appears to go to great lengths around things like the matching rules in https://tools.ietf.org/html/rfc6125#section-6.4.  Certainly it was more sophisticated than what I was building here.  Assuming it works for us, it seems more prudent to me to rework this logic.  And is along the lines of what Clayton was originally hoping for.

I'll have to rework the openshift-restclient slightly to allow for this as an option, but it is the better way to go IMO.

I'll report back once I have some results from the prototype.

gabemontero:
The prototype checks out with using okhttp3's default hostname verifier... waiting on https://github.com/openshift/openshift-restclient-java/pull/252 to merge, followed by maven publish / release cut, and then I'll update this pull to defer to okhttp3 for ssl hostname verification.

gabemontero:
Still waiting on the maven publish / release cut for the restclient (the pr there has merged), but I went ahead and pushed the resulting changes in the plugin if folks want to look.

I'll update the pom.xml once the release cut has happened.

gabemontero:
The cutting of `5.4.0.Final` of openshift-restclient-java on the jboss nexus repo apparently hit a openshift push image to docker registry network timeout :-) (https://issues.jboss.org/browse/JBIDE-23771 and https://github.com/openshift/origin/issues/12603) 

As such, I can't cite `5.4.0.Final` as a maven dependency.  We'll hold out a bit and see if they can get this resolved before falling back on using `5.4.0-SNAPSHOT` (or possibly getting them to try a `5.5.0.Final`).  We've used their SNAPSHOT's before, so have some amount of confidence in it.

I've pushed updates that currently uses `5.4.0-SNAPSHOT`.  I also tweaked the implementation based on what I saw with https://github.com/fabric8io/kubernetes-client wrt bypassing the okhttp3 hostname verifier when in skip tls mode.

gabemontero:
OK, the extended tests for this PR passed.  The job failed (as did the last PR I did) not because of some env hiccup, but because the plugin tests were reworked and the job def isn't quite right anymore ... see `The test image was changed, check for openshift/jenkins-plugin-snapshot-test on Dockerhub`.   There is in fact no such image on docker hub:

```
gmontero ~ $ docker pull docker.io/openshift/jenkins-plugin-snapshot-test
Using default tag: latest
Trying to pull repository docker.io/openshift/jenkins-plugin-snapshot-test ... 
Pulling repository docker.io/openshift/jenkins-plugin-snapshot-test
Error: image openshift/jenkins-plugin-snapshot-test not found
gmontero ~ $ 
```

Going to merge this and address the ci.openshift job def separately.

PR #127, opened 2017-01-31
use default conn TO (less aggressive)

gabemontero:


src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPluginDescriptor.java
```diff
@@ -113,7 +113,6 @@ default FormValidation doTestConnection(@QueryParameter String apiURL, @QueryPar
             
             ClientBuilder cb = new ClientBuilder(getOverride(apiURL, allOverrides)).
                     sslCertificateCallback(auth).
-                    withConnectTimeout(5, TimeUnit.SECONDS).
```
bparees, line 116: what will the timeout be now?

gabemontero: The default openshift-restclient timeout that is used in our other ClientBuilder invocations, such as https://github.com/openshift/jenkins-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPlugin.java#L147-L150

It is 2 minutes.

bparees: k

PR #130, opened 2017-02-17
advertisement for new client plugin

gabemontero:
@bparees @jupierce @csrwng FYI - ad for the new plugin in the old plugin's readme


bparees:
lgtm

PR #134, opened 2017-04-05
bugfix: error message fstring requires state

nicr9:
`EXIT_DEPLOY_VERIFY_BAD_REPLICAS_BAD` requires 4 arguments but we've only been passing 3. It seems that the missing argument is the value for `state`.

This bug causes a `MissingFormatArgumentException` if we're attempting to verify a deployment whos pods have entered a `CrashBackOffLoop` state.

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftDeploymentVerification.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftDeploymentVerification.java
@@ -137,7 +137,7 @@ public interface IOpenShiftDeploymentVerification extends ITimedOpenShiftPlugin 
                 return true;
             } else {
                 if (checkCount)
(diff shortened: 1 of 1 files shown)


bparees:
[test]

gabemontero (approved):
Yep, it sure does.  Thanks for catching this and submitting the fix @nicr9 !

We'll merge pending the test passing.

Any time frame for you @nicr9 on when you'd like this incorporated into the openshift jenkins centos images?  If so, let us know and we'll try to accommodate you.

thanks again

nicr9:
@gabemontero You're welcome =)

There's no hurry! We noticed the stacktrace while fleshing out our pipeline and it was unclear from the error message why the pipeline was failing. Once we figured it out it was because of `CrashLoopBackOff` we knew what to look for and were pretty happy.

I had some spare time on my hands so I thought I'd dig around for a fix ;)

gabemontero:
OK sounds good @nicr9 .  Even though you are not in a big hurry, I should be able to turn the crank on this within the next few days to a week.

nicr9:
Sounds great! :+1:

Issue #136, opened 2017-05-16
BUG: "Test connection" always returns "Connections successful"


livelace:
Subject. Even with a wrong password/token.

gabemontero:
Yep, I see this as well putting in a bad token.

Cross referencing with our extended test, the extended test for failure supplies a bad cluster api url to confirm test failure.  We don't have a test that supplies a bad token in jenkins_pipeline.go.

gabemontero:
Perhaps the http get of healthz/ready does not require a valid token ?!?!  I have verified we are passing in the configured (invalid) token.

I'll experiment with some other uri endpoints

gabemontero:
Yep, that appears to be the case.  I added an equivalent of `oc get is -n openshift` and that produced an auth error.

Conversely, if I go with the default token (the serviceaccount's token), the `oc get is -n openshift` succeeds.

@bparees - does rest call equivalent to `oc get is -n openshift` seem to you like a valid rest call to determine whether the supplied token carries the necessary authorization ?

bparees:
is there no generic endpoint we can hit to simply validate the token?  @liggitt @deads2k

gabemontero:
there is the "/oapi/v1/users/~" endpoint, but it is messier to access that one from the java restclient

bparees:
my worry w/ the openshift NS suggestion is the day someone decides they don't want an openshift namespace.

gabemontero:
ok, i managed to hit "/oapi/v1/users/~" with relatively little pain from the java restclient after all ... pending endpoint correction from @liggitt @deads2k let's go with that

i'll submit pull momentarily

liggitt:
`/apis` will return 401 with a bad token, and exists in both openshift and kube servers

gabemontero:
@livelace if you can try out the pre-release version of this change at https://ci.openshift.redhat.com/jenkins/view/All/job/openshift-pipeline-plugin/36/s3/download/openshift-pipeline.hpi

livelace:
@gabemontero  It works, thanks! :)

gabemontero:
On Wed, May 17, 2017 at 11:19 AM, Oleg Popov <notifications@github.com>
wrote:

> @gabemontero <https://github.com/gabemontero> It works, thanks! :)
>

Cool - thanks for testing.

PR #137, opened 2017-05-16
more strigent test needed with test connection for bad token

gabemontero:
Fixes https://github.com/openshift/jenkins-plugin/issues/136

@bparees @oatmealraisin fyi / ptal


bparees:
lgtm pending agreement from @liggitt or @deads2k about the use of that endpoint.

gabemontero:
@liggitt suggested use of `/apis` in the issue a few moments ago ... I'll circle back and switch to that.
Going to let the current merge job complete first, just to see if other items pop up.

Issue #139, opened 2017-05-18
BUG: OpenshiftExec ignores timeout 


livelace:
@gabemontero 

Hello. Yesterday I tested FR https://github.com/openshift/jenkins-plugin/issues/135, I set an environment variable ${TIMEOUT} to 10 seconds and did "sleep 30" inside OpenshiftExec, all worked as expected. Today I try to exec a heavy script and I get  a timeout error - https://paste.fedoraproject.org/paste/H5NYxUjiZtkeEuFvQh58xF5M1UNdIGYhyRLivL9gydE=/raw
I tried to set global and local timeout, but it didn't help.

PS. This morning I upgraded to 1.0.43.

gabemontero:
The error being reported here stems from the restclient/okhttp3 client timing out, where we essentially have a long running REST call.  The application of timeout in the step was not applied to the restclient/okhttp3 client timeout configuration.

We should augment our configuration of that client for this step to do that.

gabemontero:
NOTE @livelace - you'll need to set the timeout in the step's advanced tab section

livelace:
@gabemontero It's not a problem if it will work as expected. Every task usually has a dedicated timeout (different complexity).

gabemontero:
[jenkins-plugin.zip](https://github.com/openshift/jenkins-plugin/files/1011644/jenkins-plugin.zip)

ok good @livelace 

before pushing this fix to the repo, if you don't mind, I'd like for you to try the plugin I've attached here; I'm not 100% certain I can construct a long running exec close enough to what you are doing to cover everything;  this way, if we have to iterate once or twice on the change, we can expedite things by my sending you hpi files directly

gabemontero:
just unzip that file and then install the hpi file as you usually do ... thanks

livelace:
Values in advanced tab - http://fastpic.co/image/Ph0k16
Error1 -
https://paste.fedoraproject.org/paste/N9mDtmSMi6uwgwVmqn8Yel5M1UNdIGYhyRLivL9gydE=/raw
Error2 -  https://paste.fedoraproject.org/paste/AP4U~0iJjOzOZqMcvQbDC15M1UNdIGYhyRLivL9gydE=/raw

livelace:
@gabemontero I go sleep now. Feel free to attach a new version - I will check it in the morning :)

gabemontero:
[...]

PR #140, opened 2017-05-19
pass timeout to restclient/okhttp3 for exec since it might be long ru…

gabemontero:
…nning

Fixes https://github.com/openshift/jenkins-plugin/issues/139

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
@@ -1,7 +1,9 @@
 package com.openshift.jenkins.plugins.pipeline.model;
 
 import com.openshift.jenkins.plugins.pipeline.Argument;
+import com.openshift.jenkins.plugins.pipeline.Auth;
 import com.openshift.jenkins.plugins.pipeline.MessageConstants;
+import com.openshift.restclient.ClientBuilder;
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.ResourceKind;
 import com.openshift.restclient.api.capabilities.IPodExec;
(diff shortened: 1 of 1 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftExec.java
```diff
@@ -59,6 +61,39 @@ default String getDisplayName() {
...
+                    sslCertificateCallback(auth).
+                    withReadTimeout((int)timeoutInMS, TimeUnit.MILLISECONDS).
+                    withWriteTimeout((int)timeoutInMS, TimeUnit.MILLISECONDS).
+                    withConnectTimeout((int)timeoutInMS, TimeUnit.MILLISECONDS).
+                    usingToken(token).
+                    sslCertificate(getApiURL(overrides), auth.getCert());
+        }
+        
+        if (auth.useCert())
+            cb.sslCertCallbackWithDefaultHostnameVerifier(true);
```
bparees, line 89: mostly lgtm, but what was the default behavior before this change?  especially around the hostname verification, since you seem to be introducing default behavior for that here now.

gabemontero: hostname verification was done before as well - https://github.com/openshift/jenkins-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftPlugin.java#L151-L152

No the only thing was passing in the timeout to client builder

bparees: ok so before this was picking up the default implementation from IOpenShiftPlugin(which included hostname verify, but not timeout behavior) and now its defining its own default implementation.  got it.

gabemontero: Cool ... yeah if you could call `super` on a default method, Java would truly implement multiple inheritance :-)

PR #141, opened 2017-05-24
    jenkins-plugin: add retry layer/adapter for REST invocations

gabemontero:
Bug 1454485
    Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1454485
    Detailed explanation
    Implement a general retry mechanism for failing HTTP REST operations
    against the OpenShift master

@bparees @oatmealraisin FYI

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
@@ -159,4 +159,10 @@ public static final String EXIT_SERVICE_VERIFY_BAD = "\n\nExiting \"%s\" unsucce
 public static final String SOCKET_TIMEOUT = " a socket level communication timeout to \"%s\" occurred.";
 public static final String HTTP_ERR = " the HTTP level communication error \"%s\" for \"%s\" occurred.";
 
+/*
+ * These messages are centered around restclient retry
+ */
+public static final String RETRY = "\nAn exception occurred invoking a REST operation against the OpenShift master.  The operation will be retried.  Exception message \"%s\".";
+public static final String GIVE_UP_RETRY = "\nAfter a few retries, giving up invoking the REST operation against the OpenShift master.  Final exception message \"%s\".";
+
 }
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/RetryIClient.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/RetryIClient.java
@@ -1,0 +1,269 @@
+package com.openshift.jenkins.plugins.pipeline.model;
+
+import hudson.model.TaskListener;
+
+import java.net.URL;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+import com.openshift.internal.restclient.DefaultClient;
+import com.openshift.jenkins.plugins.pipeline.MessageConstants;
+import com.openshift.restclient.BadRequestException;
+import com.openshift.restclient.IClient;
+import com.openshift.restclient.IOpenShiftWatchListener;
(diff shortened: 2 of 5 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/MessageConstants.java
```diff
@@ -159,4 +159,10 @@
 public static final String SOCKET_TIMEOUT = " a socket level communication timeout to \"%s\" occurred.";
 public static final String HTTP_ERR = " the HTTP level communication error \"%s\" for \"%s\" occurred.";
 
+/*
+ * These messages are centered around restclient retry
+ */
+public static final String RETRY = "\nA execption occurred invoking a REST operation against the OpenShift master.  The operation will be retried.  Exception message \"%s\".";
```
bparees: exception.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/RetryIClient.java
```diff
@@ -0,0 +1,241 @@
...
+            } catch (Throwable t) {
+                ose = new OpenShiftException(t, "retry", call);
+                listener.getLogger().println(String.format(MessageConstants.RETRY, t.getMessage()));
+                retryCount++;
+                try {
+                    Thread.sleep(1000);
+                } catch (InterruptedException e) {
+                    ose = new OpenShiftException(e, "retry", call);
+                }
+            }
```
bparees: needs a return if we didn't catch an exception?

src/main/java/com/openshift/jenkins/plugins/pipeline/model/RetryIClient.java
```diff
@@ -0,0 +1,241 @@
...
+    private final IClient client;
+    private final TaskListener listener;
+    
+    private Object retry(Callable<Object> call) throws OpenShiftException {
+        int retryCount = 0;
+        OpenShiftException ose = null;
+        while (retryCount < MAX_RETRY) {
+            try {
+                Object o = call.call();
+                return o;
```
bparees, line 66: I assume standard failures like "permission denied" or "unknown resource" do not result in exceptions from call()?

bparees (changes requested):

src/main/java/com/openshift/jenkins/plugins/pipeline/model/RetryIClient.java
gabemontero: damn i had that in there at one time .... I was messing with `Void` vs. `void` and must have forgot to put it back ... that resulted in the test failure too.   Thanks for the catch.

src/main/java/com/openshift/jenkins/plugins/pipeline/model/RetryIClient.java
gabemontero: Another good catch ... turns out restclient has `UnauthorizedException`, `NotFoundException`, and `BadRequestException` (rc's 403, 404, and 400).  I'll check for those and *NOT* retry.

gabemontero:
Ah ... the httpGet path needs some work

gabemontero:
ok tests are clean now with some more tweaks to the error handling .. have 1 more mod I'm about to try

gabemontero:
interesting ... a concurrent mod in this run uncovered a NPE ... will address NPE, and investigate if there is a deeper issue wrt retry

gabemontero:
ah yes ... the master sends back an HTTP rc 409, which the restclient does not deal with specifically ... we already had retry for this higher up the plugin stack

IMO simply addressing the NPE in the new logic is sufficient  ... will push momentarily

gabemontero:
OK ... tests are looking good.

@bparees - any more suggestions before merging?

bparees:
lgtm

PR #143, opened 2017-06-23
suppoprt multiple certs defined in file/var

gabemontero:
per recent debug with ops team, where multiple certs store in certs file mounted in the pod

@bparees fyi

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
@@ -15,6 +15,7 @@ import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Map;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
```diff
@@ -370,14 +375,20 @@ private static InputStream getInputStreamFromDataOrFile(String data, File file)
...
     	}
-    	InputStream pemInputStream = getInputStreamFromDataOrFile(certString, caCertFile);
-		CertificateFactory certFactory = CertificateFactory.getInstance("X509");
-		X509Certificate cert = (X509Certificate) certFactory.generateCertificate(pemInputStream);
-		return cert;        
+    	InputStream pemInputStream = null;
+    	try {
+    	    pemInputStream = getInputStreamFromDataOrFile(certString, caCertFile);
+            CertificateFactory certFactory = CertificateFactory.getInstance("X509");
+            return (Collection<X509Certificate>) certFactory.generateCertificates(pemInputStream);
```
bparees: formatting

bparees (changes requested):
formatting nit and lgtm

src/main/java/com/openshift/jenkins/plugins/pipeline/Auth.java
gabemontero: should be good now thanks

bparees (approved):

PR #156, opened 2017-11-09
Change invalid function name in README

christophetd:
Cf. https://github.com/openshift/jenkins-plugin/blob/98733f5baa7efc23ff8c29d24ca40b14b2d5e329/src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftCreator.java#L71


gabemontero (approved):
thanks for catching / fixing this!

Issue #160, opened 2017-11-30
Jenkins throws an error when using the openhiftBuild with the buildConfig parameter


Torbacka:
Error message from jenkins: 

> WorkflowScript: 46: Invalid parameter "buildConfig", did you mean "bldCfg"? @ line 46, column 32.
>                    openshiftBuild(buildConfig:  #'buildname')

How to reproduce error:
- Use Jenkins 2.73.3 and OpenShift Pipeline Jenkins Plugin verson 1.0.52
- Run a pipeline with the step "openshiftBuild(buildConfig:  #'buildname')"

Solution:
I think you should rename the parameter name buildConfig to bldCfg in the README.md, but maybe i have misunderstood something. #

gabemontero:
hey @Torbacka - yeah, `buildConfig` is suppose to work.  And in fact, using the openshift sample template https://github.com/openshift/origin/blob/master/examples/jenkins/pipeline/samplepipeline.yaml, which has this in the pipeline:

```
      node('nodejs') {
          stage('build') {
            openshiftBuild(buildConfig: 'nodejs-mongodb-example', showBuildLogs: 'true')
          }
          stage('deploy') {
            openshiftDeploy(deploymentConfig: 'nodejs-mongodb-example')
          }
        }

```

works for me.

I'm at the same version of jenkins and this plugin as you.

So something else is going on.

If you want to pursue further, please provide:

- the complete pipeline script
- all of the job output
- the jenkins logs

And we can debug what is going on.

Thanks.

Torbacka:
Yeah your code works fine for me. The problem seem to be that Im using declarative pipeline. When Im running it like this I get the error.

```
pipeline {
    agent any
    stages {
      stage('build') {
        steps {
            openshiftBuild(buildConfig: 'intagiutest', showBuildLogs: 'true')
        }
      }
      stage('deploy') {
        steps {
            openshiftDeploy(deploymentConfig: 'nodejs-mongodb-example')
        }
      }
    }
}
```

gabemontero:
Yep, that is the difference @Torbacka 

With declarative, the workflow engine is not longer calling https://github.com/openshift/jenkins-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/dsl/OpenShiftBuilder.java#L120

which is where we are doing the param name mapping 

A README update is in fact warranted, but rather then a generic "only use ###" it needs to be qualified around declarative vs. scripted pipelines.

[...]

PR #161, opened 2017-11-30
Removed unused parameter buildConfig

Torbacka:
#160


gabemontero:
Thanks, but see https://github.com/openshift/jenkins-plugin/issues/160#issuecomment-349028176

This is not an update we want to make.

PR #162, opened 2017-12-07
add declarative pipeline restriction note on useable param names

gabemontero:
Fixes https://github.com/openshift/jenkins-plugin/issues/160

also adds a note about IS SCMs vs. ICTs with pipeline BCs 

@openshift/sig-developer-experience fyi


gabemontero:
just a readme change, but i'll wait for any input on the verbiage before merging

bparees:
lgtm

PR #167, opened 2018-08-02
do not abort obj create/delete if we hit a master that does not suppo…

gabemontero:
…rt /swaggerapi/oapi

@openshift/sig-developer-experience fyi

to account for removal of /swapperapi/oapi REST endpoint

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftApiObjHandler.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftApiObjHandler.java
@@ -20,8 +20,16 @@ public interface IOpenShiftApiObjHandler extends IOpenShiftPlugin {
     default String fetchApiJsonFromApiServer(boolean chatty,
             TaskListener listener, Map<String, String> overrides,
             String apiDomain) {
-        return httpGet(chatty, listener, overrides, this.getApiURL(overrides)
+        String json = null;
+        try {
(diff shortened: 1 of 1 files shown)


src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftApiObjHandler.java
```diff
@@ -20,8 +20,16 @@
...
+        String json = null;
+        try {
+        	json = httpGet(chatty, listener, overrides, this.getApiURL(overrides)
                 + "/swaggerapi" + apiDomain + "/v1");
+        } catch (Throwable t) {
+        	if (chatty) {
+        		listener.getLogger().println(t.getMessage());
+        	}
+        }
+        return json;
```
bparees, line 32: caller can handle null?

gabemontero: yes

gabemontero: https://github.com/openshift/jenkins-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/model/IOpenShiftApiObjHandler.java#L27-L30

gabemontero:
hmm perhaps this seldom updated repo is not fully wired for merging PRs ... or there was a hiccup when the tests passed

merging manually

PR #168, opened 2018-08-03
Eol warning new discovery

gabemontero:
@openshift/sig-developer-experience fyi / ptal


bparees:
lgtm.

@bmcelvee can we make sure we get something in the v3.11 release notes warning users that v3.11 will be the last release supported by the jenkins openshift pipeline plugin and they will need to move to the openshift client plugin with the next release?

@gabemontero we should also update this to say it's deprecated and the upcoming v3.11 release will be the last version it works with:
https://docs.openshift.org/latest/using_images/other_images/jenkins.html#pipeline-plug-in

gabemontero:
docs pr https://github.com/openshift/openshift-docs/pull/11369

PR #172, opened 2019-06-11
convert mvn/plugin repos from http to https

gabemontero:
/assign @bparees 
/assign @adambkaplan


gabemontero:
v1.0.56 of the old plugin has been initiated at the update center ... 3.11 update only

PR #177, opened 2020-03-23
Update .travis.yml

akram:


akram:
/assign @gabemontero

cannot find explicit doc on setting openjdk8 with travis, but this may fix the build.

akram:
hooray, it looks better, and it builds successfully.
@gabemontero can you LGTM it ?
so the other ones cand be merged also

